PR #17, opened 2015-02-10
Add support for skipping or expecting failure on a test

cdent:
This is needed because during the test writing process there
are often _many_ things that ought to succeed but don't that are
beyond the scope of the immediate project. For example a framework
not properly returning a 405 when a method is not allowed, or
returning 405 but then not returning the Allow header.

This feature required switching the base test case from unittest to
testtools. The former has changed (in Python 3.4) the internals of how
expecting failure is handled and as a result making it work with
dynamically generated tests was proving extremely difficult.

An additional shell driven test is provided for coverage of
skip and fail.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -13,7 +13,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""A single HTTP request reprensented as a subclass of ``unittest.TestCase``
+"""A single HTTP request represented as a subclass of ``unittest.TestCase``
 
 The test case encapsulates the request headers and body and expected
 response headers and body. When the test is run an HTTP request is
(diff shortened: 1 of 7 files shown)


cdent:
@FND if you've got a moment a sanity check on this would be helpful, but no worries if not.

gabbi/case.py
```diff
@@ -13,23 +13,40 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""A single HTTP request reprensented as a ``unittest.TestCase``
+"""A single HTTP request reprensented as a subclass of ``unittest.TestCase``
```
FND: typo: "represented"

FND:
seems sensible to me - though of course I lack familiarity to provide a truly informed assessment

(I was confused when yesterday I said "shouldn't the testing framework be responsible for this"; I was thinking of `pytest -x`)

PR #21, opened 2015-03-02
Allow templating headers

sileht:
Hi, 

This change allows to use template in the request headers.

Cheers,


cdent:
Simple enough, have you got a particular use case? I just want to know so I can properly update the docs/tests/etcs.

sileht:
I would like to set the keystone token like that line 4:
https://review.openstack.org/#/c/159538/4/gnocchi/tests/gabbi/gabbits_gate/base.yaml

cdent:
Ah, yeah, of course. I'll push out a new release in a few minutes.

PR #22, opened 2015-03-10
Remove useless usage of lambda

jd:


cdent:
Here's the replication strategy for why this needs to be there:

* go to your gnocchi master
* `tox -r -epy27-gabbi`
* edit `.tox/py27-gabbi/lib/python2.7/site-packages/gabbi/fixture.py` to remove the dreaded `lambda`
* `tox -epy27-gabbi`

For me, at least, that's a fail. I'm pretty sure it was in the process getting gabbi to work with gnocchi that the lambda arrived. I hate it too because it introduces yet another piece of indirection at a level where there's always been too much indirection anyway (paste app factories, direct wsgi callables, application symbols in modules etc etc etc) but that's what it took to get it to work _and_ make it work with other setups.

Not ideal, obviously, because what's being done here is me throwing up my hands and saying "I have no idea what's really going on but this makes it work" but sometimes that's just the way it goes.

Unless you've got additional input, I reckon we can probably close this for now and revisit the issue when we are all smarter?

cdent:
Hmm. I may be completely wrong on this. I've taken another pass at it and will have new data in a moment.

cdent:
Ah, yeah, you're totally right. Sigh, clearly something went amiss in my too-manual testing.

/me bows to @jd

cdent:
This is of course obvious upon reflection.

To go faster, slow down.

PR #27, opened 2015-03-31
attempted to clarify response handler documentation

FND:
just a few suggestions, not sure whether they make more sense than what was there before

also, I'm not sure I'm using `response_strings` right; the docs seem to disagree with existing samples

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -1,33 +1,42 @@
 Response Handlers
 =================
 
-Responses handlers are code which control the evaluation of the
-HTTP response in each test. For example the `response_strings` key
-described in :doc:`format` is a response handler. In many cases the
-default handlers described there will be sufficient for creating
-useful tests. When they are not, it is possible to add custom
-handlers by subclassing :class:`~gabbi.handlers.ResponseHandler` and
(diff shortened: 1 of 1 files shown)


docs/source/handlers.rst
```diff
@@ -1,33 +1,42 @@
...
-  against the HTTP response. A loop is made across each entry in the
-  test key with data passed to action:
-  If ``test_key_value`` is a list, then the arguments to``action`` are (in order):
-
-  * ``self``: The current object.
+* ``test_key_value``: The key's default value, either an empty list or empty
+  dict. It is also a class level value.
+* ``action``: An instance method which tests the expected values
+  against the HTTP response - it is invoked for each entry, with the parameters
+  depending on the default value. If ``test_key_value`` is a list, then the
```
cdent: If we is "if this is a list" then we have these args, then there must be a "if this is a dict" then we have these args at the same level, you can't bury the if a dict in the "then we have" part.

If on the other hand you say, there are the args but if it is a list this one isn't used, it reflects reality and has grammatical parallelism.

cdent:
Gonna go ahead and merge this in and then edit for taste.

PR #29, opened 2015-04-08
Wrap the http request to trap WSGIAppError and reraise

cdent:
To create a strong boundary between the client and the server,
wsgi-intercept wraps exceptions raised in the intercepted app in a
WSGIAppError. In most situations this is a good thing. However if
gabbi is being used to do TDD against a nascent API this behavior
can obscure useful traceback information when there is a
catastrophic and untrapped exception in the server code.

To deal with this the WSGIAppError is caught and the contained
exception is reraised in all its glory to spew into STDERR for
debugging purposes.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -27,7 +27,9 @@ import os
 import re
 import sys
 
+import wsgi_intercept
 import jsonpath_rw
+import six
 from six.moves.urllib import parse as urlparse
 from testtools import testcase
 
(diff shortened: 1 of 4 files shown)


cdent:
Would be handy to get some feedback from @FND and @pkilambi on this. Thanks.

gabbi/simple_wsgi.py
```diff
@@ -51,6 +51,9 @@ def __call__(self, environ, start_response):
             ('X-Gabbi-url', request_url),
         ]
 
+        if request_method == 'DIE':
+            raise Exception('because you asked me to')
```
FND: I'm assuming this wasn't supposed to be committed?

gabbi/case.py
```diff
@@ -27,7 +27,9 @@
 import re
 import sys
 
+import wsgi_intercept
```
FND: merely out of curiosity: why not `from wsgi_intercept import WSGIAppError`?

cdent: That's a habit from openstack land where one is obliged to only import modules. I'll adjust it to non openstack styles

gabbi/simple_wsgi.py
cdent: No it was. There are various features in gabbi that are a bit hard to properly test from gabbi itself, so rather than actually test them the simple_wsgi app has ways to ask that certain pieces of the code are traversed. The line above combined with the test at line 109 in self.yaml insure that the exception handling is at least encountered.

gabbi/case.py
cdent: Actually apparently I'm using that style quite a bit throughout so may as well stick with it.

PR #30, opened 2015-04-15
Add a command line to to run yaml-tests from stdin.

cdent:
A console-script reads stdin and generates a TestSuite to be run.
This can be used as a nice way to test running services without any
test-building-in-python requirements.


cdent:
This might be useful, allows this kind of thing:

   http://paste.openstack.org/show/203965/

I'm a bit unsure on the usual things I'm unsure about: docs, level of formalizing. I kind of prefer to leave that until things have fleshed out.

/cc @FND

PR #34, opened 2015-06-04
Preliminary Getting Started suggestions for running the tests

kiminoa:
Small doc change to add a few hints for newbies on running the tests.


cdent:
Thanks very much for this. I'll do some tweaks from the base you've provided.

PR #37, opened 2015-06-10
JSONPath formatting breadcrumb in the docs

kiminoa:
In case newbs besides me don't grok the whole JSONPath is a thing thing.

### DIFF
--- a/docs/source/example.yaml
+++ b/docs/source/example.yaml
@@ -167,7 +167,8 @@ tests:
 # "$ENVIRON" is a pseudo dictionary providing access to the current
 # environment.
 # "$RESPONSE" provides access to the JSON in the prior response, via
-# JSONPath.
+# JSONPath. See http://goessner.net/articles/JsonPath/ for JSONPath 
+# formatting.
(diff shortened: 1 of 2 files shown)


cdent:
Great. Only comment/question I would have is: Gabbi's implementation is built upon jsonpath-rw and there are some things in the URL you linked to that are not in jsonpath-rw and some things in jsonpath-rw that are not in the blog post. Does this matter? If it does, does that mean we need more links and more info where those links are?

My initial thinking is "not really" because the basics are close enough and that's what we're really concerned with here, we don't want to be encouraging people to be doing backflips in their JSONPath.

What do you think?

kiminoa:
Doh, I didn't appreciate the distinction.

http://jsonpath-rw.readthedocs.org/en/latest/ is probably the better URL... but I did find the JSONPath url to be a quick skim-n-grok whereas I had to hunker down and read the jsonpath-rw documentation. Still, the right solution in my mind is to be exact. I'll switch the URL unless you object.

Thanks for your patience! :)

cdent:
Seems good thanks!

PR #47, opened 2015-06-17
Build tests prefix

cdent:
A prefix is applied to all the unqualifed URLs in a suite of tests.
This is useful for testing live applications that are mounted at
different points under a wsgi server.


pkilambi:
:+1: This looks useful

jasonamyers:
:+1:

PR #56, opened 2015-07-16
Ensure response header comparisons are case-insensitive

### LINKED ISSUE #55: case-insensitive headers
As far as I can tell, gabbi enforces lowercase headers:

```yaml
response_headers:
    content-type: text/html; charset=utf-8
```

```
... ✓ front page returns HTML
```

vs.

```yaml
response_headers:
    Content-Type: text/html; charset=utf-8
```

```
... E front page returns HTML

ERROR: front page returns HTML
	 "'Content-Type' header not available in response keys: dict_keys(['server', 'content-type', 'access-control-allow-origin', 'date', 'access-control-allow-credentials', 'connection', 'content-location', 'content-length', 'status'])"
```

From my perspective, the second version is more readable - so it would be nice if header name comparisons were case-insensitive.

FND:
fixes #55 

I'm not sure reusing `handler` and `self.test` in the test is appropriate, but I wanted to avoid repetitive boilerplate (for readability)

### DIFF
--- a/gabbi/handlers.py
+++ b/gabbi/handlers.py
@@ -120,6 +120,8 @@ class HeadersResponseHandler(ResponseHandler):
     test_key_value = {}
 
     def action(self, test, header, value):
+        header = header.lower()  # case-insensitive comparison
+
         response = test.response
         header_value = test.replace_template(value)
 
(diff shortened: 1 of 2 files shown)


cdent:
Yup that's all fine, thanks.

PR #70, opened 2015-07-30
Get rid of testtools and adjust gabbi.report accordingly

cdent:
testtools has too much magic, making it difficult to have direct access to test error information. Having access to that test error information is required to make the output from gabbi-run useful.

The change attempts to replicate the existing behavior of gabbi-run but add no more. That should be for later efforts.

@FND if you're able to confirm the "existing behavior" part then we can probably merge.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -26,11 +26,12 @@ import os
 import re
 import sys
 import time
+import unittest
+from unittest import case
 
 import jsonpath_rw
 import six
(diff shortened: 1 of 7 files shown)


FND:
:+1: lgtm :ship:

:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

cdent:
:rage4: thanks

PR #74, opened 2015-08-02
NETLOC shouldn't have default port

sileht:
Hi,

If scheme in http and port 80, NETLOC shouldn't have port set.

Same for https

Cheers,

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -252,7 +252,9 @@ class HTTPTestCase(unittest.TestCase):
         netloc = self.host
 
         if not url_scheme:
-            if self.port:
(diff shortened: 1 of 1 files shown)


cdent:
Yup makes sense. Initially I was thinking perhaps self.port should just become unset earlier in the process, but that makes it harder to do weird things like have non-ssl on the use ssl port or vice versa (which somebody might want to do).

cdent:
I did a followup commit with some tests c673770a962d6ed287979d5cce9fb046a6b29eeb and decided to make your port comparisons be happy with either ints or str.

This will be released early this afternoon.

PR #75, opened 2015-08-02
Add query_parameters field

cdent:
This adds a query_parameters field which can be used to add query
parameters to the url of the test in which it is used. It is a dict in
a form suitable for being passed to the urlencode() method.

This change is cumbersome because urlencoded in python2 and 3 have
different behaviors: In python2 a utf-8 encoded string is required.
In 3 it will figure out the right thing. So we here we just encode.

We also need to deal with the fact that a numeral in YAML will be
provided to Python as a numeric value and we need to stringify that in
a version independent fashion.

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -39,6 +39,9 @@ Many of these items allow substitutions (explained below).
   fully qualified URL (with host and scheme). If not qualified the
   test builder will be responsible for determining host and scheme.
   **Required**
+* ``query_parameters``: An optional dictionary of query parameters
+  that will be added to the ``url``. If there is an existing set of
+  query parameters they wil be extended.
 * ``method``: The request method to use. Defaults to ``GET``.
 * ``status``: The expected response status code. The default is
   ``200``. If necessary you may indicate multiple response codes
@@ -111,6 +114,7 @@ character must be used at both ends.
 All of these variables may be used in all of the following fields:
 
 * ``url``
+* ``query_parameters``
 * ``data``
 * ``request_headers``
 * ``response_strings``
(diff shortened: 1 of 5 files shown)


cdent:
@FND you'll need to review this to see if it suits requirements and is implemented correctly and completely. It was harder to get right that I though it would be. There are probably ways to make the code better. Please do if you like.

Also, if you feel it is appropriate, add an example to example.yaml in the docs.

And any other comments you might feel compelled to do.

docs/source/format.rst
```diff
@@ -39,6 +39,9 @@ Many of these items allow substitutions (explained below).
   fully qualified URL (with host and scheme). If not qualified the
   test builder will be responsible for determining host and scheme.
   **Required**
+* ``query_parameters``: An optional dictionary of query parameters
+  that will be added to the ``url``. If there is an existing set of
+  query parameters they wil be extended.
```
sileht: Should be a list of tuple, the query parameter can be provided multiple times.

cdent: It's a dict like this (which is one the data structures that urlencode accepts and is the easier one to represent in yaml):

```yaml
query_parameters:
    foo: hello
    bar:
         - 1
         - 2
```

which will result in `foo=hello&bar=1&bar=2`.

I agree it is fairly confusing, thus the invitation to @FND to extend this to fix the docs as he sees fit (it was his feature request).

FND: I'm sorry that I still haven't gotten around to following up - might still take a few days though.

Nevertheless: I don't think the structure above is confusing at all, it's in line with how many libraries handle query parameters. Of course the dictionary discards order, thus `foo=hello&bar=1&bar=2` is equivalent to `bar=1&bar=2&foo=hello` - but that was deemed acceptable. (Of course order for multiple values of the same parameter is retained.)

sileht: OK, the format looks fine, I have missed the underlying list.

cdent:
merged by hand

PR #76, opened 2015-08-04
Add `len` extention to the json parser

sileht:
This extends the jsonpath_rw parser with customisation
for gabbi.

This first one is `len`

### DIFF
--- a/gabbi/json_parser.py
+++ b/gabbi/json_parser.py
@@ -1,0 +1,56 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import jsonpath_rw
+
+
+class Len(jsonpath_rw.JSONPath):
+    """The JSONPath referring to the len of the current object.
+
+    Concrete syntax is '`len`'.
+    """
+
+    def find(self, datum):
+        datum = jsonpath_rw.DatumInContext.wrap(datum)
+        try:
+            value = len(datum.value)
+        except TypeError:
+            return []
+        else:
+            return [jsonpath_rw.DatumInContext(value,
+                                               context=None,
+                                               path=Len())]
+
+    def __eq__(self, other):
+        return isinstance(other, Len)
+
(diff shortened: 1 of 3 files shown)


gabbi/json_parser.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Len(jsonpath_rw.JSONPath):
+    """The JSONPath referring to the len of the current object.
+
+    Concrete syntax is '`len`'.
+    """
+
+    def find(self, datum):
+        datum = jsonpath_rw.DatumInContext.wrap(datum)
+        return [jsonpath_rw.DatumInContext(len(datum.value),
```
cdent: I'm not sure if it should, but does it make any sense for it also to work on str and dict?

In fact, would it make any sense to just call len and handle TypeError if the datum.value can't len()?

I can certainly imagine situations where people will want to know the length of a dict that is being returned.

What do you think?

Bringing in @FND as he may have an opinion too. FND, in case you feel like this is out of the blue, especially with regard to discussions we've had about dealing with location, there's some precedent, I've been looking for months to [add sorts](https://github.com/kennknowles/python-jsonpath-rw/pull/25) but hadn't figured out the best way to integrate. This gives us a place to hang such things. _If_ we choose to do so.

FND: I don't have the brains to parse the code in detail right now or think deeply about the conceptual implications. However your suggestion WRT type agnosticism makes sense. In fact, I can imagine users simply assume that it works for any iterable, so not supporting that would likely end up being surprising.

gabbi/json_parser.py
```diff
@@ -0,0 +1,51 @@
...
+        return isinstance(other, Len)
+
+    def __str__(self):
+        return '`len`'
+
+    def __repr__(self):
+        return 'Len()'
+
+
+class GabbiJsonPathParser(jsonpath_rw.parser.JsonPathParser):
```
FND: I'd greatly prefer using "JSON" rather than "Json". That should also be more consistent (cf. `jsonpath_rw.JSONPath`).

cdent: In this case the subclass ought to replicate the naming standard established by the super? So while `JSONPath` is used elsewhere, since the parser uses `JsonPathParser` ...?

gabbi/json_parser.py
sileht: I make it generic and handle TypeError, str len is clearly a must have.

gabbi/json_parser.py
FND: Ah, right, I missed that (shouldn't have tried without suitable brains... ). Ugh, but yeah, fair enough.

cdent:
There was some irc-based discussion about the possibility of this being a module of its own, an independent package of jsonpath improvements that gabbi happens to use. I think there's merit in that idea but for now we'll go ahead and merge this and consider it as an option for the future.

PR #77, opened 2015-08-05
Add more debug information for JSONResponseHandler

sileht:
hi,

When the expected value doesn't match the received
for the JSONResponseHandler, this change adds the received
value into the assertion error message.

Cheers,


cdent:
ayup, thanks

Issue #15, opened 2015-02-05
when status does not match expected, we should probably show body


cdent:
Something like

```python
self.assertIn(observed_status, statii, 'status %s is not in %s: %s' % (observed_status, statii, output))
```

where `output` is the current response body, which will need to be passed to the method.

cdent:
Not actually that certain about this.  Could get long.

In general the failure reporting is far from ideal.

cdent:
#79 pretty much covered this

PR #79, opened 2015-08-05
Add more information when status assertion fail

sileht:
Hi,

Having the response output is useful to understand
why the status code doesn't match the expected one.

Cheers,

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -36,6 +36,8 @@ import wsgi_intercept
 from gabbi import json_parser
 from gabbi import utils
 
+MAX_CHARS_OUTPUT = 2000
+
 REPLACERS = [
     'SCHEME',
     'NETLOC',
@@ -308,11 +310,13 @@ class HTTPTestCase(unittest.TestCase):
 
         # Decode and store response
         decoded_output = utils.decode_content(response, content)
-        content_type = response.get('content-type', '').lower()
+        self.content_type = response.get('content-type', '').lower()
         if (decoded_output and
-                ('application/json' in content_type or
-                 '+json' in content_type)):
+                ('application/json' in self.content_type or
+                 '+json' in self.content_type)):
             self.json_data = json.loads(decoded_output)
+        else:
+            self.json_data = None
         self.output = decoded_output
 
     def _run_test(self):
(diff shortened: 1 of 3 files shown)


cdent:
There's an old issue #15 for this and on there I comment about concerns about the response body being potentially quite long and that being very annoying, so I dithered and haven't done anything yet.

Are the length concerns legit or should we just not worry?

Would it make sense to only include the response when the content-type indicates something non-binary?

sileht:
I see, I will add something with content-type and length

sileht:
I have made a new proposal it limits the output by lines, but I could moves to number of char.

cdent:
This is bit hairy, I hope that @FND and @jasonamyers will be able to comment too.

gabbi/case.py
```diff
@@ -402,7 +406,34 @@ def _test_status(self, expected_status, observed_status):
             statii = [stat.strip() for stat in expected_status.split('||')]
         else:
             statii = [expected_status.strip()]
-        self.assertIn(observed_status, statii)
+
+        self.assert_in_or_print_output(observed_status, statii)
+
+    def assert_in_or_print_output(self, expected, iterable):
+        if self.content_type != "application/binary":
```
cdent: (I made this comment already but it seems to have been swallowed somewhere, so if you've seen this already, my apologies.)

Maybe use [gabbi.utils:not_binary](https://github.com/cdent/gabbi/blob/master/gabbi/utils.py#L82) here so that things like `image/png` also get skipped.

gabbi/case.py
```diff
@@ -402,7 +406,34 @@ def _test_status(self, expected_status, observed_status):
...
+                msg = "'%s' not found in %s%s" % (
+                    expected, response,
+                    '\n...truncated...' if is_truncated else ''
+                )
+            else:
+                msg = "'%s' not found in %s, %sresponse:\n%s" % (
+                    expected, iterable,
+                    'truncated ' if is_truncated else '',
+                    response)
+            self.assertIn(expected, iterable, msg)
```
cdent: I'm probably prematurely optimizing but it seems like perhaps msg shouldn't be created until we know there has been a failure otherwise all that effort to produce a nice message will have gone to waste.

jasonamyers:
I'd prefer to limit chars instead of lines. lines are kinda arbitrary.

gabbi/case.py
sileht: Done

gabbi/case.py
sileht: Done

cdent:
ok

PR #83, opened 2015-08-13
Make user agent identifier more informative

FND:
there doesn't seem to be much of a suitable standard, but this seems
closer to common patterns

cf. https://en.wikipedia.org/wiki/User_agent#Format_for_human-operated_web_browsers ff.


cdent:
ya, dat better

PR #84, opened 2015-08-13
colorized verbose output

FND:
see commit message

I'm not thrilled about the colors, but wasn't sure it's worth trying  `Style.DIM` with `MAGENTA` (which would require extending `get_colorizer`) to get a purple for headers

screen shot 2015-08-13 at 19 40 37

with `s/YELLOW/GREEN/`:

screen shot 2015-08-13 at 19 57 26


FND:
resubmitted against the proper branch in #85

PR #88, opened 2015-09-03
Gabbi run argparse

cdent:
Add support for using argparse with gabbi-run so that it produces some reasonable help and can take a URL as the target (instead of host[:port] [/prefix]).

/cc @FND, @jasonamyers

### DIFF
--- a/docs/source/runner.rst
+++ b/docs/source/runner.rst
@@ -15,14 +15,22 @@ line::
 
     gabbi-run [host[:port]] < /my/test.yaml
 
+or::
+
+    gabbi-run http://host:port < /my/test.yaml
+
 To facilitate using the same tests against the same application mounted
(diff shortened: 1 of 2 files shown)


jasonamyers:
:+1:

docs/source/runner.rst
```diff
@@ -15,14 +15,22 @@ line::
...
+
+    gabbi-run http://host:port/prefix < /my/test.yaml
+
 The value of prefix will be prepended to the path portion of URLs that
 are not fully qualified.
 
-If a ``-x`` argument is provided then ``gabbi-run`` will "fail fast"
-exiting after the first test failure.
+If a ``-x`` or ``--failfast`` argument is provided then ``gabbi-run`` will
+"fail fast", exiting after the first test failure.
```
FND: this addition seems somewhat redundant, why not simply "... will exit ..."

PR #89, opened 2015-09-07
added optional shortcut for specifying HTTP method and URL

FND:
before:

```yaml
method: GET
url: /foo
```

after:

```yaml
GET: /foo
```

Quoting from an earlier e-mail:

> The particular shortcut I proposed (`<verb>: <URL>`) would make the most
> fundamental part of an HTTP interaction more concise - plus it would
> more closely resemble the actual HTTP syntax.
>
> That's not just less to type, it might also improve readability (less
> repetitive noise) - assuming you're reasonably familiar with HTTP and
> its conventions (e.g. less common verbs like PATCH or OPTIONS might not
> be immediately recognizable as such, though the all-uppercase aspect is
> a decent hint).

----

⚠ work in progress

* [x] implementation
* [x] proper tests
* [x] documentation

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -3,13 +3,20 @@ Test Format
 
 Gabbi tests are expressed in YAML containing an HTTP request and an
 expected response. Each YAML file is an ordered sequence of requests.
-The bare minimum YAML file for a single request is::
+A minimal YAML file for a single request is::
 
     tests:
        - name: the name of a test
+         GET: /
+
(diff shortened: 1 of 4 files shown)


cdent:
This mode was approved by acclaim at pyconuk 2015 so if we get some docs going its a goer.

FND:
So I [read](http://anticdent.org/pyconuk-day-one.html). :tada: It's on my list for next week.

PR #90, opened 2015-09-08
cli: target must be optional

sileht:
Since last release of gabbi enforce to set a target
on the command line, when it was not required before.

This have breaks the ceilometer integration tests that
query multiples API endpoints and don't set this target.

This change fixes that.


cdent:
Nice catch, thanks.

PR #95, opened 2015-10-01
added support for custom response handlers to gabbi-run

FND:
the CLI option should be self-explanatory - we'll probably need a little more guidance for extension authors eventually, but for now the referenced function's docstring should suffice


cdent:
@jasonamyers if you get a chance to glance at this that would be awesome but no biggie if not, @FND and I are both happy with it.

jasonamyers:
I like the idea and the test pass so :+1:

PR #97, opened 2015-10-08
Extract parse url

cdent:
Extract _parse_url to its own utility method to allow it to be cleaned and for the query_parameters test element to be handled in its own method.

_parse_url was high complex according to the [radon](https://pypi.python.org/pypi/radon) tool.

### DIFF
--- a/gabbi/tests/test_utils.py
+++ b/gabbi/tests/test_utils.py
@@ -84,7 +84,7 @@ class ExtractContentTypeTest(unittest.TestCase):
         self.assertEqual('utf-8', charset)
 
 
-class ColorozieTest(unittest.TestCase):
+class ColorizeTest(unittest.TestCase):
 
     def test__colorize_missing_color(self):
         """Make sure that choosing a non-existent color is safe."""
@@ -95,7 +95,7 @@ class ColorozieTest(unittest.TestCase):
         self.assertNotEqual('hello', message)
 
 
-class CreateUrlTest(unittest.TestCase):
+class CreateURLTest(unittest.TestCase):
 
     def test_create_url_simple(self):
         url = utils.create_url('/foo/bar', 'test.host.com')
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -63,7 +63,7 @@ class TestBuilder(type):
 
 def build_tests(path, loader, host=None, port=8001, intercept=None,
                 test_loader_name=None, fixture_module=None,
-                response_handlers=None, prefix=None):
+                response_handlers=None, prefix=''):
     """Read YAML files from a directory to create tests.
 
     Each YAML file represents an ordered sequence of HTTP requests.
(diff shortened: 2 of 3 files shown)


cdent:
@FND and @jasonamyers at your leisure etc

gabbi/tests/test_utils.py
```diff
@@ -80,6 +83,9 @@ def test_extract_content_type_bad_params(self):
         self.assertEqual('text/colorful', content_type)
         self.assertEqual('utf-8', charset)
 
+
+class ColorozieTest(unittest.TestCase):
```
FND: non parlo italiano

gabbi/tests/test_utils.py
```diff
@@ -87,3 +93,42 @@ def test__colorize_missing_color(self):
 
         message = utils._colorize('BLUE', 'hello')
         self.assertNotEqual('hello', message)
+
+
+class CreateUrlTest(unittest.TestCase):
```
FND: I wish you were consistent in your use of capitals (where consistent here means `CreateURLTest`, because I said so).

cdent: Okay, but in my mind it is your way that is inconsistent: initial caps, _even if_ it is an acronym.

gabbi/tests/test_utils.py
cdent: well done

gabbi/case.py
```diff
@@ -437,6 +412,31 @@ def _test_status(self, expected_status, observed_status):
 
         self.assert_in_or_print_output(observed_status, statii)
 
+    def _update_query_params(self, original_query_string, query_params):
+        """Update a query string from query_params dict.
+
+        An OrderedDict is used to allow easier testing and greater
```
FND: I always feel bad when adjusting runtime code merely for the sake of testing. I suppose this is necessary because you (quite reasonably) want to use string comparisons in the tests below, rather than parsing the query string and verifying the individual parts thereof?

gabbi/tests/test_utils.py
FND: You crazy, bro!

gabbi/tests/test_utils.py
```diff
@@ -80,10 +83,52 @@ def test_extract_content_type_bad_params(self):
         self.assertEqual('text/colorful', content_type)
         self.assertEqual('utf-8', charset)
 
-    def test__colorize_missing_color(self):
+
+class ColorizeTest(unittest.TestCase):
+
+    def test_colorize_missing_color(self):
```
cdent: @FND, I'm not going to bother fixing this, but the fact that there was a dunder here was intentional.

FND: Oh, why's that?

cdent: On Fri, 9 Oct 2015, FND wrote:

> Oh, why's that?

The method being tested was _colorize.

-- 
Chris Dent                                   http://burningchrome.com/
                                 [...]

PR #99, opened 2015-10-10
Move gabbit tests into the tests dir

cdent:
Originally there was no tests dir. When unit-like tests were added
they were put in a newly created tests directory. Since it is there
may as well use it.


cdent:
@FND, I was there so I went ahead and did this, was pretty much just a question of synchronicity.

FND:
:+1: j'approuve

PR #100, opened 2015-10-11
removed misleading comment and spurious line breaks

FND:
see commit message

also, I remain confused by your seemingly arbitrary line length limits

### DIFF
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -39,6 +39,7 @@ from gabbi import handlers
 from gabbi import httpclient
 from gabbi import suite as gabbi_suite
 
+
 RESPONSE_HANDLERS = [
     handlers.HeadersResponseHandler,
     handlers.StringResponseHandler,
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -261,11 +261,8 @@ def test_suite_from_yaml(loader, test_base_name, test_yaml, test_directory,
     try:
         test_data = test_yaml['tests']
     except KeyError:
-        raise GabbiFormatError(
-            'malformed test file, "tests" key required')
+        raise GabbiFormatError('malformed test file, "tests" key required')
     except TypeError:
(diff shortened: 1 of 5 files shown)


gabbi/driver.py
```diff
@@ -261,11 +262,8 @@ def test_suite_from_yaml(loader, test_base_name, test_yaml, test_directory,
     try:
         test_data = test_yaml['tests']
     except KeyError:
-        raise GabbiFormatError(
-            'malformed test file, "tests" key required')
+        raise GabbiFormatError('malformed test file, "tests" key required')
     except TypeError:
-        # Swallow this exception as displaying it does not shine a
```
cdent: Please don't remove this.

You're welcome to improve it (or the message it is commenting upon), but it is providing a useful piece of information:

* The previous KeyError is caused by the 'tests' key not existing and we express exactly that in the error message
* The TypeError is caused by test_yaml not being a dict (for whatever reason) and I couldn't think of a clean or concise way to say "we expected this but got this" without doing a bit of extra work so I left the comment to excuse the sin of raising a new exception without reference to the exception that inspired it (and that is very much a sin).

cdent:
Neither the pep8 command not the [pep itself](https://www.python.org/dev/peps/pep-0008/#blank-lines) say that module level variables should get double leading blanks, so what's up with those changes?

On line length limits: As you know I tend to prefer narrow code, for sake of my ability to read it so I will often break strings that get beyond the 60s. There are also frequently cases where code which was somewhere else at a different indent gets moved to a lesser indent and is not reflowed. I don't know which happened in this case but the usual comment/thought remains: At least I wrote the code.

FND:
> module level variables should get double leading blanks

It's more about imports being followed by double blanks.

However, it appears you're right; this seems to be a convention I picked up on top of PEP8. Nevertheless, that convention is being followed elsewhere in gabbi, so at least these adjustments provide consistency.

I'll revisit the deleted comment and reopen.

PR #101, opened 2015-10-11
stylistic tweaks

FND:
see commit messages for details

recreated based on discussions in #100; personally, I'm not convinced that the new, verbose comment is better than not having any comment at all - but at least it's no longer confusing (IMHO yadda yadda)

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -38,6 +38,7 @@ from gabbi import __version__
 from gabbi import json_parser
 from gabbi import utils
 
+
 MAX_CHARS_OUTPUT = 2000
 
 REPLACERS = [
@@ -48,7 +49,6 @@ REPLACERS = [
     'HEADERS',
     'RESPONSE',
 ]
-
 
 # Empty test from which all others inherit
 BASE_TEST = {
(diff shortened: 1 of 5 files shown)


cdent:
I think this is fine but since you seem to not be entirely confident, some feedback:

* If you're not happy with the change to the comment (I think it is an improvement) just don't make any change to that chunk of code: In a set of of commits that is otherwise about whitespace changing a comment would be considered out of place by some (I don't mind, I'm happy to get improvements of any kind).
* I'm not aware of imports requiring bounded 2 blank lines. It's not mentioned [here](https://www.python.org/dev/peps/pep-0008/#blank-lines) whereas 2 blank line bounding is mentioned for other things. However, it's a fine change if you like it.

I'm happy to merge this as is if you are.

FND:
As I tried to convey in the OP, I still think this comment is better than the previous one - so I might not be happy (shock!), but it's an improvement nonetheless.

As for bunching this up in a single PR, that's mostly for historical reasons: #100 started out as what seemed like a trivial change, then spiraled out of proportion (again... though that did result in some useful clarifications I think). Separating these commits seemed like needless bureaucracy.

My rationale for the blank lines are explained here: https://github.com/cdent/gabbi/pull/100#issuecomment-147186115

> it appears you're right; this seems to be a convention I picked up on top of PEP8. Nevertheless, that convention is being followed elsewhere in gabbi, so at least these adjustments provide consistency.

I just went ahead and merged it myself. Will step AFK now; blue skies.

Issue #94, opened 2015-10-01
Add hypothesis example to docs or wiki


cdent:
See http://wildfish.com/blog/2015/10/01/using-gabbi-and-hypothesis-test-django-apis/

PR #102, opened 2015-10-11
attempted to clarify test and suite generation

FND:
This is very much a light version of what I had planned [originally](https://github.com/cdent/gabbi/commit/415d76c2cfaffe04b9379bbe76af9eaff29ec6cb#commitcomment-13701857): I was considering introducing a set of intermediate classes (`GabbiTestDescriptor[Collection]`?) to encapsulate and shuffle around some of the logic in both `test_suite_from_yaml` and `TestMaker`, but I'm no longer convinced it's worth the effort (for now anyway).

Nevertheless, these leftovers of that aborted effort seem to better describe what's going on, if only by grouping related aspects within the existing function.

What still irks me is that `test_suite_from_yaml` actually doesn't have anything to do with YAML: It's merely processing a dictionary that once happened to originate from a YAML file. However, I couldn't figure out how to get around that without introducing new terminology (see above).

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -50,7 +50,7 @@ REPLACERS = [
     'RESPONSE',
 ]
 
-# Empty test from which all others inherit
+# Basic test template determining both valid keys and default values
 BASE_TEST = {
     'name': '',
     'desc': '',
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -257,8 +257,6 @@ def test_update(orig_dict, new_dict):
 def test_suite_from_yaml(loader, test_base_name, test_yaml, test_directory,
                          host, port, fixture_module, intercept, prefix=''):
     """Generate a TestSuite from YAML-sourced data."""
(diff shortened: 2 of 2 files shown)


cdent:
Yeah, this seems like a good bit of tidying.

If you've seen my [recent 2.0 related blog posting](http://anticdent.org/planning-gabbi-20.html) then you will see your thoughts "to encapsulate and shuffle around some of the logic" reflected there.

The lack of yaml in test_suite_from_yaml irks me too. It's probably safe to rename it as it is not really a part of the public api[1], which would bring things closer to something easier to abstract and indirect.

[1] The public api is effectively:

* the format of the tests
* the functionality of the things that will eventually be in ContentHandlers:
    * response handlers
    * replacers
    * `data` to request bodies
* the build_test method
* the arguments of and inputs to gabbi-run

If we do rename `test_suite_from_yaml` we might want to leave behind a shim to protect people who have done their own integrations like the hypothesis + gabbi thing linked from #94 or at least tell those people about the change.

PR #109, opened 2015-12-09
Add response_forbidden_headers handler

### LINKED ISSUE #108: Unable to express header not being present in response
When attempting to translate some python tests to gabbi I (@cdent) was unable to translate a test that was asserting that a particular header was _not_ present in the response.

It's currently impossible to make these kinds of negative assertions (not just of headers, but that's the case that came up) and it's likely we need to make it possible.

Syntax is a confusing issue:

* The current value for `response_headers` is a dict of key (header) value pairs. If we want to say "this header is not present" that doesn't fit in that structure.
* If we want to express this thing (e.g. header) is present but does _not have value X_ that is semantically different from expressing not being present. We need to think about both when designing to avoid suboptimal grammar for both.

cdent:
After much discussion this was determined to be the easiest and
cleanest way to accomplish a test which asserts "this header must
not be present". The ResponseHandler framework is used to good
effect to make short code.

The value of the response_forbidden_headers is a list of headers.
Case will be normalized to lower, which matches the internal
representation of the response headers dictionary.

Fixes #108

### DIFF
--- a/docs/source/example.yaml
+++ b/docs/source/example.yaml
@@ -93,6 +93,14 @@ tests:
       response_headers:
           content-type: /application/json/
 
+# If a header must not be present in a response at all that can be
+# expressed in a test as follows.
+
+    - name: test forbidden headers
+      url: /resource
+      response_forbidden_headers:
+          - x-special-header
+
 # All of the above requests have defaulted to a "GET" method. When
 # using "POST", "PUT" or "PATCH", the "data" key provides the
 # request body.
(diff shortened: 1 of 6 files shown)


cdent:
@jasonamyers or @FND if one of you could put your baleful eye on this that would be great, I'd like to release tonight. If you can't, no problem, I'm confident enough with the implementation. It's mostly a syntax sanity check I'm after.

jasonamyers:
:+1:

PR #111, opened 2016-01-14
Remove tox from test-requirements

hayderimran7:
tox must be pre-installed in order to run tests, as `tox -epep8` or simply `tox` .  
Adding tox to test requirements is redundant as these packages are only required if used in `gabbi/tests` folder.


cdent:
Thanks!

PR #115, opened 2016-03-05
Jsonpath more

### LINKED ISSUE #86: considere integrating python-jsonpathw-ext
@sileht made https://pypi.python.org/pypi/jsonpath-rw-ext and it's rather useful

### LINKED ISSUE #114: asserting JSON array contents in arbitrary order
It's not unusual for JSON APIs to return arrays whose order is irrelevant (i.e effectively sets, e.g. [JSON API](http://jsonapi.org/format/)'s `data` object graph).

Testing for entries within such arrays seems tricky. In order to assert that a specific entry exists, I'm currently resorting to this:

```yaml
response_json_paths: # XXX: order currently hard-coded, but not actually relevant
    $.data[0].id: login
    $.data[0].links.self: http://example.org/login
    $.data[1].id: profiles
    $.data[1].links.self: http://example.org/profiles
    $.data[2].id: posts
    $.data[2].links.self: http://example.org/posts
```

This is brittle because order isn't guaranteed. The only alternative seems to be to use filter expressions:

```yaml
$.data[?(@.attributes.id = "login")]
```

That's not particularly readable though.

Is this something #86 could help with, perhaps via `sorted`?

cdent:
This adds support for sorting and filtering in json paths, using jsonpath_rw_ext with a slight modification to allow the results to be traversable (see [pull request](https://github.com/sileht/python-jsonpath-rw-ext/pull/7)).

Fixes #114
Fixes #86 

/cc @FND @jasonamyers

### DIFF
--- a/docs/source/jsonpath.rst
+++ b/docs/source/jsonpath.rst
@@ -7,10 +7,14 @@ validating response bodies that are formatted as JSON and making
 reference to that JSON in subsequent queries. `jsonpath_rw`_ is used
 to process the JSONPath expressions.
 
-To address a common requirement when evaluting JSON responses, an
-extension has been made to the default implementation of JSONPath.
-This extension is ``len`` and will return the length of the current
-datum in the JSONPath expression.
+To address common requirements when evaluting JSON responses,
+extensions have been made to the default implementation of JSONPath
+using `jsonpath_rw_ext`. It adds three useful functions:
+
+#. Return the length of the current datum using ``len``.
+#. Sort the current datum using ``sorted`` and ``[/name]``.
+#. Filter using ``[?name = "cow"]`` to select an item in the
+   current datum.
 
 Here is a simple JSONPath example, including use of ``len``. Given JSON data
 as follows::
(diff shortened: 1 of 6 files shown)


jasonamyers:
I think this looks good!

docs/source/jsonpath.rst
```diff
@@ -36,6 +40,8 @@ follows::
         # The string at beta has five chars
         $.beta.`len`: 5
 
-There are more JSONPath examples in :doc:`example`.
```
jasonamyers: Should we add examples of sorting and filtering?

cdent: Well, yes, but I ran out of energy (only made this change because I needed it for something else). Thus the links below to external sources. There are some examples in the tests too. If you've got the cycles to add something that's awesome, otherwise we can do it later.

docs/source/jsonpath.rst
```diff
@@ -7,10 +7,14 @@ validating response bodies that are formatted as JSON and making
...
-To address a common requirement when evaluting JSON responses, an
-extension has been made to the default implementation of JSONPath.
-This extension is ``len`` and will return the length of the current
-datum in the JSONPath expression.
+To address common requirements when evaluting JSON responses,
+extensions have been made to the default implementation of JSONPath
+using `jsonpath_rw_ext`. It adds three useful functions:
+
+#. Return the length of the current datum using ``len``.
+#. Sort the current datum using ``sorted`` and ``[/name]``.
```
FND: FWIW, I only understood this by looking at the tests. (I'm aware of https://github.com/cdent/gabbi/pull/115#discussion_r55124235 though.)

cdent: Yeah, I would like to make the docs for all of the json path features much more complete but I didn't want needing that to hold up the release (I needed the feature working before this coming wednesday). It's probably time for another robust run through the docs. I may have time for that now that I'm past the crunch time for nova (which resulted in utter failure in anything getting merged).

PR #117, opened 2016-03-10
Switch to using jsonpath_rw_ext 1.0.0

cdent:
Version 1.0.0 merges the non-monkeypatch version of the changes
which this patch removes from gabbi.


cdent:
If this is cool, I'll release it as 1.14.0

PR #119, opened 2016-03-14
Adjust header evaluation to always use strings

### LINKED ISSUE #112: If the test value for a response header is not treated by yaml as a string there is an error
```
File "/Users/cdent/src/enamel/.tox/functional/lib/python2.7/site-packages/gabbi/handlers.py", line 149, in action
        if header_value.startswith('/') and header_value.endswith('/'):
    AttributeError: 'float' object has no attribute 'startswith'
```

when:

```yaml
    - name: default version
      desc: no header sent
      GET: /
      request_headers:
        content-type: application/json
      response_headers:
        vary: /openstack-enamel-api-version/
        openstack-enamel-api-version: 0.1
```

cdent:
Sometimes yaml will provide a header response value as a numeral.
We don't want that. So to be explicit both sides of the header value
test are stringified: both the value in the real reaponse, and the
value of the expected response.

Fixes #112 

/cc @FND, @jasonamyers

### DIFF
--- a/gabbi/handlers.py
+++ b/gabbi/handlers.py
@@ -128,6 +128,8 @@ class HeadersResponseHandler(ResponseHandler):
 
     If a header value is wrapped in ``/`` it is treated as a raw
     regular expression.
+
+    Headers values are always treated as strings.
     """
 
     test_key_suffix = 'headers'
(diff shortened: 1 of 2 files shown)


jasonamyers:
:+1:

PR #121, opened 2016-03-18
Switch to use urllib3 from httplib2

cdent:
httplib2's maintenance status is in doubt so using urllib3 is a bit
more sane because it is the active and preferred choice for http
library that provides a high level of control and low level of magic.

This changes both the client for making the requests, and the
wsgi-intercept handling.

The details of the changes are in gabbi.httpclient which creates
a lightweight facade over the urllib3.PoolManager.request method
to transform the results into something that has the shape of an
httplib2 response, content pair.

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -17,12 +17,32 @@ from __future__ import print_function
 import os
 import sys
 
-import httplib2
+import urllib3
 
 from gabbi import utils
 
 
-class VerboseHttp(httplib2.Http):
+class Http(urllib3.PoolManager):
+
+    def request(self, absolute_uri, method, body, headers, redirect):
+
+        retries = 1 if redirect else False
+        response = super(Http, self).request(
+            method, absolute_uri, body=body, headers=headers,
+            redirect=redirect, retries=retries)
+
+        # Transform response into something akin to httplib2
+        # response object.
+        content = response.data
+        status = response.status
+        reason = response.reason
+        headers = response.headers
+        headers['status'] = str(status)
+        headers['reason'] = reason
+        return headers, content
+
+
+class VerboseHttp(Http):
     """A subclass of Http that verbosely reports on activity.
 
     If the output is a tty or ``GABBI_FORCE_COLOR`` is set in the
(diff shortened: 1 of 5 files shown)


cdent:
First pass at switching gabbi to use urllib3 instead of httplib2. This works but may be missing some nuance. Review appreciated from @FND, @jasonamyers and @sigmavirus42.

sigmavirus24:
Also pinging @lukasa for giggles. ;)

gabbi/httpclient.py
```diff
@@ -17,12 +17,32 @@
...
 
 from gabbi import utils
 
 
-class VerboseHttp(httplib2.Http):
+class Http(urllib3.PoolManager):
+
+    def request(self, absolute_uri, method, body, headers, redirect):
+
+        retries = 1 if redirect else False
```
sigmavirus24: I assume this is retries for a redirect? If so, you might want to be a bit more explicit like:

```
retries = False
if redirect:
    retries = urllib3.util.Retry(total=False, redirect=1)
```

Otherwise, that will do `Retry.from_int` which, if I am remembering correctly, sets the total number of retries. I don't know if you care about connection errors and other things being retried instead of redirects.

sigmavirus24:
I just have the one note. Otherwise, things look fine to me.

gabbi/httpclient.py
cdent: The ideal end goal I'm after is:

* if there is a 302, 303, etc and redirect is True, follow it, otherwise return that status code, don't follow
* in all cases, if there is any error, don't do a retry and force the "real" error to be raised

Is that what your example does. If not, how do I do that in a good clean way?

Lukasa: The urllib3 Retry class lets you do all of that. It's remarkably configurable: [take a look](https://urllib3.readthedocs.org/en/latest/helpers.html#urllib3.util.retry.Retry).

Lukasa:
So this looks reasonable to me too. One thing: it may be worth seeing how effectively you can "encourage" @elarson to work with the urllib3 team to try to get his caching stuff into urllib3.

gabbi/httpclient.py
sigmavirus24: Hm, let me answer your question with a question:

- So how many 302/303s do you want to follow?

> in all cases, if there is any error, don't do a retry and force the "real" error to be raised

So by default, I think urllib3 will try a maximum of 10 (over all things it retries) retries. I think you can disable that elegantly by passing `retries=False` but I've forgotten and haven't used that in a while. I'll need a few minutes to confirm.

If you don't care about following a potentially large number of redirects, you should be able to just say `redirects=True` and be done with all of it.

My example above was allowing for exactly 1 redirect and otherwise returning the status code of the second response (in the event of a redirect) and raising errors otherwise.

sigmavirus24:
I had plans for that too but have had far more pressing priorities outside of work since this isn't a work/OpenStack priority for anyone.

PR #122, opened 2016-04-06
Pytest

cdent:
This branch provides rudimentary support for running gabbi managed tests with py.test as well as the tests-of-gabbi with py.test in additional tox targets. There is a shortcoming (noted in the docs) but I don't think it is enough of a problem to not make this available for the hardy and willing.

/cc @FND @jasonamyers

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,8 +13,9 @@ more details on features and formats.
 
 Gabbi is tested with Python 2.7, 3.4, 3.5 and pypy.
 
-Tests can be run using `unittest`_ style test runners or from the
-command line with a `gabbi-run`_ script.
+Tests can be run using `unittest`_ style test runners, `pytest`_ (with
+one limitation described in the docs) or from the command line with
+a `gabbi-run`_ script.
 
 There is a `gabbi-demo`_ repository which provides a tutorial via
 its commit history. The demo builds a simple API using gabbi to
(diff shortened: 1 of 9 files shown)


jasonamyers:
:+1: I like this, and I don't think the pytest caveats are that bad.

README.rst
```diff
@@ -13,8 +13,9 @@ more details on features and formats.
 
 Gabbi is tested with Python 2.7, 3.4, 3.5 and pypy.
 
-Tests can be run using `unittest`_ style test runners or from the
-command line with a `gabbi-run`_ script.
+Tests can be run using `unittest`_ style test runners, `pytest`_ (with
+one limitation described in the docs) or from the command line with
```
FND: It's not clear which docs you're referring to here?

docs/source/example.rst
```diff
@@ -14,21 +14,24 @@ harness.
 Loader
 ------
 
-Test Loader
-~~~~~~~~~~~
+To run the tests they must be generated in some fashion and then
+run. This is accomplished by a test loader. Initially gabbi only
```
FND: History is irrelevant to users?

docs/source/example.rst
```diff
@@ -54,5 +57,27 @@ See the `source distribution`_ and `the tutorial repo`_ for more
...
+   :language: python
+
+This can then be run with the usual pytest commands. For example::
+
+   py.test -svx pytest-example.py
+
+.. warning:: When using the unittest runners, it is possible to select
+             just one test from a YAML file and cause it and all its
+             prior tests to run. This **does not** work when using
+             pytest. It will traverse the prior tests and then exit
```
FND: I'm not sure I understand what "traverse" means in this context.

README.rst
cdent: fixed, patch on its way

docs/source/example.rst
cdent: I think it is as it explains why things are so crufty.

docs/source/example.rst
cdent: I'll attempt to clarify.

cdent:
For sake of history, note that this didn't quite work and was fixed by #126

PR #129, opened 2016-04-16
move pytest from test-req to req

zqfan:
we already import pytest in our production code, hence it should be
in requirements.txt rather than test-requirements.txt, otherwise
downstream may fail when usig gabbi

ref: http://logs.openstack.org/53/306753/2/check/gate-aodh-dsvm-functional-hbase/a8f48e6/console.html#_2016-04-16_15_53_52_203

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 pbr
 six
+pytest
 PyYAML
 urllib3>=1.11.0
 jsonpath-rw-ext>=1.0.0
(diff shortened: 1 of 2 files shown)


cdent:
Thanks, I saw this email and thought it was an issue not a pull request, so just fixed it in aa257c9fcd9f9778ae9b60

A new release is being created right now, will be out as 1.17.2 [now](https://pypi.python.org/pypi/gabbi).

PR #132, opened 2016-05-05
First pass at implementing $LAST_URL feature

cdent:
If $LAST_URL is used anywhere a replacer is accepted, it will
be replaced with the URL from the previous test.

At the moment the URL is the full url that is passed into the http
client. Thus it includes the scheme, the netloc and any query
parameters that may have been added by the use of the
`query_parameters` test key.

It's not certain that this is the right choice. The other option is
to use the value of `url` that is in the test input. The is what's
actually the value of the url or {GET,POST,..} key in the test.
This is problematic though as there are at least two ways this could
be interpreted:

* the literal string that is in test['url'] in the test data
  structure
* that string after replacers have been run

Using the final full URL seems to suit the use cases I can think of
and feels somehow less ambiguous so that's what I implemented, but
it too has some issues:

* If $LAST_URL has query parameters in it, trying to _replace_ (not
  extend) those parametes via `query_parameters` doesn't really work.
  There's a failing test to demonstrate this problem.

(there may be others I haven't thought of yet)

Input desired from @FND, @jd  and others please.


FND:
In an [IRC conversation](http://p.anticdent.org/2HCs), @cdent and I arrived at the conclusion that adhering to the Principle of Least Surprise, it's probably best to force the distinction between `url` and `query_parameters`. That is, `$LAST_URL` should be expanded (i.e. replacers already applied), but _not_ include a query string - if the latter is required, explicit `query_parameters` should be supplied (duplicated from the previous test if necessary).

cdent:
In [further conversation](http://p.anticdent.org/3KVS) we agreed that stripping query string is bad.

So the summary is that $LAST_URL will be: the value of test_data['url'] used in the last request after and replacers ($RESPONSE, $LOCATION etc) have been processed but without query_parameters processing and without host expansion.

Unless @jd has further comments that change this idea, I'll implement it soon so we can get this out of the way and move on to content handlers.

cdent:
@jasonamyers Unless you have something to say about this I'm gonna merge it up soon.

jasonamyers:
Looking at it now

jd:
I like the least surprise solution. Removing the query parameters would be a surprise to me.

And if there's a way to use $LAST_URL + adding query parameter I think you got most things covered correctly.

jasonamyers:
I don't really view the query parameters as part of the URL, so I didn't really find it surprising like @jd did. However, I do think he makes a good point that should be either handled via documentation that states no params or via code that offers both paths.  I prefer the doc route as I want a clean URL each time without the params

cdent:
Thanks very much for the comments

PR #135, opened 2016-05-31
Cookie substitution

cdent:
This branch provides a way of using a $COOKIE substitution to send the cookies set via a `set-cookie` in the prior response in this current request. Both @FND and I have gazed over it and been a bit "well yeah, it does what it is supposed to do but somehow seems meh" so it would be great to get some further input from @jasonamyers and @martynbristow on this, just for sanity checking.

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -211,6 +211,9 @@ processed in the order given.
   string value of the environment variable is ``"True"`` or
   ``"False"`` then the resulting value will be the corresponding
   boolean, not a string.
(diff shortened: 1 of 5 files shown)


jasonamyers:
will review tonight, not at a computer for a bit

jasonamyers:
👍  I'm okay with this

cdent:
Thanks @jasonamyers gonna push this out into the world so we can see what people think of it.

cdent:
https://pypi.python.org/pypi/gabbi/1.20.0

PR #137, opened 2016-06-01
Get gabbi working with IPv6 addresses

### LINKED ISSUE #136: gabbi can't cope with an ipv6 host argument
If an ipv6 address is passed in host when doing `build_tests` the built url is incorrect. An ipv6 address in a url is supposed to be bracketed: `[::1]`.

cdent:
According to rfc 2732 an IPv6 address in a url needs to be enclosed
in [] (e.g. http://[::1]/foobar). Gabbi was not doing this. Instead
it was passing host information directly to its URL generation
routine and just using it. When doing tests against real servers
running on ipv6, this results in failures.

The changes here try to do the bare minumum to get the right results
without changing the create_url method too much. It's likely that in
some future version of gabbi we will want to change a lot here (the
current orientation towards hosts and ports instead of URLs is the
result of the wsgi-intercept integration) but for now we want to
keep the change limited.

I hate myself for the mocks in test_runner. A clear sign of bad
abstraction in runner.py, however this is not the place to be
changing that. Another item for the later list.

Fixes #136

### DIFF
--- a/docs/source/runner.rst
+++ b/docs/source/runner.rst
@@ -32,5 +32,8 @@ or in the target URL::
 The value of prefix will be prepended to the path portion of URLs that
 are not fully qualified.
 
+Anywhere host is used, if it is a raw IPV6 address it should be
+wrapped in ``[`` and ``]``.
+
 If a ``-x`` or ``--failfast`` argument is provided then ``gabbi-run`` will
 exit after the first test failure.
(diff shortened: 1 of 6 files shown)


cdent:
If @FND could look for sanity and @jd, or @sileht  to see if it will address the gnocchi+gabbi+tempest issue that would be great.

EmilienM:
:+1:

jd:
lgtm

Issue #138, opened 2016-06-02
when SSL is enabled, 400 Bad Request


EmilienM:
Testing OpenStack Gnocchi Tempest tests, that run Gabbi, on SSL & IPV6 network, the Gabbi tests fail with 400 Bad Request

More details: http://logs.openstack.org/61/318961/8/check/gate-puppet-openstack-integration-3-scenario001-tempest-centos-7/5bac108/console.html#_2016-06-01_22_29_31_718

EmilienM:
@cdent feel free to close this one if duplicated with another SSL bug. Not sure it was related though.

cdent:
This is probably related to #50 and #105 but I'll take a look. For now the fix is likely in the gnocchi tempests themselves.

I assume it also doesn't work for SSL & IPV4?

cdent:
#139 has the fix for this, but it will require a change in the gnocchi tests to make it work

Where it [determines the port](https://github.com/openstack/gnocchi/blob/04aab41d76b85c59f0b6a9de2bb80046ac6a1da6/gnocchi/tempest/scenario/__init__.py#L46) it will also need to decide a True or False value for `require_ssl` and then pass `require_ssl=require_ssl` as one of the args to build tests. I can do those changes once we've had a chat about it.

But only after #139 merges and is released.

cdent:
@EmilienM the related [gnocchi fix](https://review.openstack.org/#/c/324337/).

Issue #105, opened 2015-10-18
default to standard SSL port for HTTPS URIs


FND:
$ gabbi-run https://api.github.com < sample.yaml

results in "[Errno 61] Connection refused", whereas

    $ gabbi-run https://api.github.com:443 < sample.yaml

works succeeds.

It seems appropriate to default to port 443 for HTTPS URIs.

cdent:
Originally this was done to force people to be explicit about wanting ssl. With the advent of gabbi-run, the way it does targeting, and various other refactorings, clearly something has become confused and lost since then.

So yeah, there's definitely room for improvement here.

cdent:
This has some interesting issues. It is easy enough to get the port set to 443 at the runner level, however, when we get around to actually doing the request, since `test_data['ssl'] = False` (the default), the URL is created with `http`.

Because httplib2 has this code:

```python
            (scheme, authority, request_uri, defrag_uri) = urlnorm(uri)
            domain_port = authority.split(":")[0:2]
            if len(domain_port) == 2 and domain_port[1] == '443' and scheme == 'http':
                scheme = 'https'
                authority = domain_port[0]
```

everything works. But this is an implementation detail that we should not rely on.

So I'm thinking the underlying error/issue is the way in which SSL is being managed (some of which is discussed in #50). Changing that would be a fairly fundamental thing, unless we do something like manage the test base somehow.

Ideas?

FND:
As agreed on IRC, the solution will likely be to set `defaults['ssl'] = True` somewhere.

cdent:
Based on the existence of #138 I've looked into this a bit more and I think what can be done is in the small number of entry points to test_suite_from_dict we can directly manipulate the value of defaults in the dict that is passed. More details to come shortly, but it is likely this will be slightly tidier if we merge the runner-cleanup branch first.

Issue #50, opened 2015-06-25
In 'live' testing scenarios argument passing to build_tests is convoluted and SSL may not work


cdent:
If you want to use `build_tests` to create real TestCases against a live server it's likely you know the URL and that would be most convenient thing to pass instead of having to parse out the host, port and prefix (script_name) and then pass those.

In addition, if you have a URL you know if your server is SSL but the tests may not have been written to do SSL (with an `ssl: true` entry). Because of the test building process this is a bit awkward at the moment. It would be better to be able to say "yeah, this is SSL"  for the whole run.

cdent:
See also #105 which has related ssl things to say.

I suppose one way to do this is to add yet another argument (`url` or `uri`) which is mutually exclusive of `host`, `port`, `prefix` and `intercept`.

Such a lot of arguments though...

cdent:
There's still part of this which is not done, which is just passing the url to gabbi instead of parsing beforehand.

cdent:
One option for doing this that wouldn't need to extend the method signatures even more would be to have something like a `url_to_kwargs` method that returned a dict of `host`, `port`, `require_ssl` and `prefix` that could be passed straight into `built_tests`.  Of course now that I've written that down it's probably just easier to pass url and if it is present it overrides all of those.

Opinions?

PR #139, opened 2016-06-02
Allowing forcing SSL in build_tests and gabbi-run

cdent:
build_tests gains a require_ssl argument which, if set to True,
makes all the loaded test suites default to 'ssl: True'.

gabbi-run will interpret a target containing 'https' as meaning
that the tests in the provided yaml should default to 'ssl: True'.

Fixes: #50
Fixes: #105
Fixes: #138

The changes here are the naive basics to get the desired behavior.
There's an existing cleanup branch on which we can clean this up
later or if people are able to come in for a review soon can happen here.

/cc @FND, @jd, @EmilienM, @jasonamyers


jd:
LGTM

PR #152, opened 2016-06-10
Add a pytest plugin to fix fixture handling

### LINKED ISSUE #123: running prior tests when selecting just one test from a yaml file does not work with pytest
See: https://gabbi.readthedocs.org/en/latest/example.html#pytest

and: http://lists.idyll.org/pipermail/testing-in-python/2016-April/006716.html

### LINKED ISSUE #128: pytest test count is inaccurate
The stop and start methods of the TestSuite (which makes sure fixtures are started) are currently counted as tests, so for each yaml file there are N+2 tests.

Fixing this probably requires becoming properly acquainted with the pytest test collection and fixture handling and not using the current form of yielding tests.

cdent:
This is a backwards compatible improvement to the way gabbi works
with pytest. Tests are loaded (and yielded in the same way) but
they are filtered more correctly, and fixture start and stops are
done more correctly.

This allows the test count to be accurate and test selection (using
-k) to work properly.

Fixes #128
Fixes #123

The details of this change are more clear with an understanding of
how pytest plugins and hooks work, but the gist is that a plugin has
been created and then forcibly installed. The plugin traverse the
collected tests to find and filter out the two tests associated with
each suite that start and stop fixtures. These are attached as
attributes to the first and last selected (if any) tests in the
relevant suite. When a test is asked to run:

* if it has a 'starter', the start will run
* then the test will run itself
* it has a prior that has not run, that will be run
* as the tests run if a test with a 'stopper' is found, the stopper
  will be run

### DIFF
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -103,6 +103,11 @@ def py_test_generator(test_dir, host=None, port=8001, intercept=None,
     This uses build_tests to create TestCases and then yields them in
     a way that pytest can handle.
     """
+
+    import pytest
+    pluginmanager = pytest.config.pluginmanager
+    pluginmanager.import_plugin('gabbi.pytester')
+
     loader = unittest.TestLoader()
     result = reporter.PyTestResult()
     tests = build_tests(test_dir, loader, host=host, port=port,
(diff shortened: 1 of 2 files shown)


cdent:
@jasonamyers and @josdotso this is probably relevant for the stuff you are working on, if you could take a look that would be great

josdotso:
@cdent Sorry I didn't get a chance to look at this during this week.  I've been traveling for training.  I'm going to keep this in my inbox to eventually take a look though.  Thanks!

cdent:
No worries @josdotso. I've gone ahead and merged it and it is 1.23.0. My testing _seemed_ to indicate that it is working correctly. If you do get a chance to check it out and it is busted please let me know asap.

I'll be travelling too, early next week, but mostly around otherwise.

PR #153, opened 2016-06-12
Enable a 'url' arg to build_tests

### LINKED ISSUE #50: In 'live' testing scenarios argument passing to build_tests is convoluted and SSL may not work

cdent:
The value of url overrides host, port and prefix if it is set.
This will make some live testing scenarios more straightforward.

Fixes #50


cdent:
@jd and @EmilienM, this good for you guys?

EmilienM:
:+1:

PR #157, opened 2016-06-16
Send a warning when a yaml filename has an underscore

### LINKED ISSUE #144: Q: What characters are legal / recommended for test name in YAML?
I suspect plain alphanum + spaces is recommended, but I might sometimes want to do hyphen or parens.  So, just being thorough.

Thanks!

cdent:
TestSuites are grouped by test runners by the name that is
generated for the tests within. To ease that grouping it is
useful to have filenames with a known set of characters.

Grouping is also used for py.test fixture handling.

This change raises a warning if a filename has an '_' in it,
effectively declaring that though it is possible to use '_',
it can impact grouping.

Fixes #144


cdent:
As demonstrated by https://review.openstack.org/#/q/topic:cd/fix-gabbi-1.22 the grouping is fragile so we should do what we can to make it at least predictable.

PR #161, opened 2016-07-13
Add retype_numerals parameter to replace_template

### LINKED ISSUE #147: substitutions result in string which can be a problem for JSON
If using jsonschema on the server side and expecting integers for some values, using a $RESPONSE style substitution in data in a POST or PUT can fail because a value may appear as a string not an int.

Will need to check if the type of the value is preserved and it is just the replace regex that is doing the breakage. If so, it ought to be fixable with relative ease.

### LINKED ISSUE #150: Poll and count don't use substitutions
Discussion was started here: https://github.com/cdent/gabbi/issues/144#issuecomment-224988014

**my initial ask:**

> p.s. rather than opening another issue immediately, another quick question: Is it possible to source integers from environment variables in the YAML? I ask because I just tried to source poll delay and count from environ, but test failed complaining about strings vs. integers. I tried unquoting the value during shell export, but no dice. Worth opening a new issue?

**@cdent's reponse from #144:**

> ...However, poll and count, don't use substitutions so even if the environment was producing stuff it wouldn't get pulled, so some fixes would be needed in the poll handling.
> Yeah, if you could make a fresh issue, that would be great.

cdent:
When replace_template is used against the value of a test key there are
contexts within which if the value looks like a number it should be treated
as a number. For example when creating JSON from the `data` key.

A retype_numerals parameter has been added to replace_template which,
when true, will attempt to turn something that looks like a float or int into
a float or int.

Fixes #147 
Fixes #150

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -139,7 +139,7 @@ class HTTPTestCase(unittest.TestCase):
             self.prior.run(result.TestResult())
         self._run_test()
 
-    def replace_template(self, message, make_numbers=False):
+    def replace_template(self, message, retype_numerals=False):
         """Replace magic strings in message."""
         if isinstance(message, dict):
             for k in message:
@@ -162,7 +162,7 @@ class HTTPTestCase(unittest.TestCase):
                         raise AssertionError(
                             'unable to replace %s in %s, data unavailable: %s'
                             % (template, message, exc))
-                    if make_numbers:
+                    if retype_numerals:
                         try:
                             if '.' in message:
                                 message = float(message)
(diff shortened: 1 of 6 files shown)


cdent:
This is not ready to go, is just here for discussion. See #147 for some of the reason why something like this may be required.

cdent:
@josdotso and @FND if you have time to look and think about this, that would be great. It is addressing some of the problems.

gabbi/case.py
```diff
@@ -419,7 +437,8 @@ def _test_data_to_string(self, data, content_type):
                 else:
                     return info
         else:
-            data = json.dumps(data)
+            data = self.replace_template(data, retype_numerals=True)
```
FND: How do we know that numeric values is what we want here? (Though I suppose it doesn't actually hurt either way.)

josdotso:
@cdent 

Looks good to me.  If I understand correctly, among other things, this will allow environment variables to be used to specify polling count and delay at runtime.  Sounds good!

gabbi/case.py
cdent: The reason for the reordering here is so that when data is passed to replace_template it is a dict or list instead of string. Since replace_template is recursive, each chunk of data will be processed in a small way, so that only chunks which are just a number will become numbers.

However you've just caused me to find a possible bug, or a total flaw in the logic, so there will be at least one more commit.

cdent:
I think this needs to be abandoned in favor of a new approach that integrates with the refactored content handlers.

PR #163, opened 2016-08-16
Add the framework for including a FAQ in docs

cdent:
Add a faq.rst and two simple questions to get the ball rolling
on having a gabbi faq in the docs.

This format probably won't scale in the long run, but is useful in
the short run and while the format might become lame the content
will not.

/cc @FND @jasonamyers

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -1,0 +1,38 @@
+
+Frequently Asked Questions
+==========================
+
+.. note:: This section provides a collection of questions with
+          answers that don't otherwise fit in the rest of the
+          documentation. If something is missing, please create an
+          issue_.
+
+          As this document grows it will gain a more refined
+          structure.
+
+General
+~~~~~~~
+
+Is gabbi only for testing Python-based APIs?
+--------------------------------------------
+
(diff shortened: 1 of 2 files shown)


jasonamyers:
YES PLEASE 👍

josdotso:
This is good.  BTW, I was ramping up on Gabbi until our white box testing became black box testing, meaning the suite I was writing is slated to go beyond the API.  I still like Gabbi though!  Hope to return to it soon.  Cheers!

cdent:
> our white box testing became black box testing, meaning the suite I was writing is slated to go beyond the API

 I should probably add a faq on that ;)

In openstack the gabbi tests are often intermixed with other functional tests, so there's no clear separation of black and white box tests.

Or to put it another: Keep using gabbi! ;)

josdotso:
Ah.  That's brilliant.  I'll have to take a bit to think it through to better comprehend the tactical aspects but that is definitely cool.  I am planning to do something like that for Selenium web driver tests (keep them in the same tree executing using testtools (unittest-based), tox, testrepository).  Thanks!

PR #169, opened 2016-09-06
Add a custom move to six to make mock available for both Python 2 and 3.

### LINKED ISSUE #168: bare "mock" imported even for Python 3
Since the `mock` package is available in the Python 3 stdlib as `unittest.mock`, `mock` is not always available independently for Python 3.  Its import should be protected in some way, either with a try/except block or through an addition to six.moves.

dhduvall:
Fixes #168


cdent:
That's nice and tidy, thanks.

PR #172, opened 2016-09-20
Add some doc and FAQ about quieting pytest warnings

cdent:
As it will take some time and effort to replace the existing yield
tests, add some hopefully helpful information on how to deal with
the warnings.

/cc @FND

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -21,6 +21,16 @@ Is gabbi only for testing Python-based APIs?
 No, you can use :doc:`gabbi-run <runner>` to test an HTTP service
 built in any programming language.
 
+Workarounds
+~~~~~~~~~~~
+
+pytest produces warnings about yield tests. Can I make them stop?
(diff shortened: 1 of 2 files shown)


docs/source/faq.rst
```diff
@@ -21,6 +21,16 @@ Is gabbi only for testing Python-based APIs?
 No, you can use :doc:`gabbi-run <runner>` to test an HTTP service
 built in any programming language.
 
+Workarounds
+~~~~~~~~~~~
+
+pytest produces warnings about yield tests. Can I make them stop?
```
FND: If _Workarounds_ above is an H2, this should be an H3.

cdent: The issue is actually in the next header. rst set header levels by what it finds first, so the rest of the document indicates that === is h1 ~~~~ is h2 and --------- is h3, so the header starting 'pytest' is wrong. If you're going to fix the other thing, if you could fix this, that would be great.

cdent: Oh never mind, I see, you're saying exactly what I'm saying, I just haven't adapted to where you put the comment. I would have put it the line below.

Anyway, fix it please, or at least make a bug, this has already merged.

Issue #104, opened 2015-10-16
gabbi-run should support multiple input files


FND:
It doesn't make much sense that `gabbi-run` currently only supports a single suite (via STDIN).

Ideally, it should be possible to pass multiple YAML files as positional arguments:

```
$ gabbi-run ... -- foo.yaml bar.yaml *.yaml
```

However, that's tricky because `target` and `prefix` are optional positional arguments. Thus we have no reasonable way to determine whether a positional argument is an input file (or do we?) - so we might have to resort to a named option (e.g. `-f`).

(According to @cdent, maintaining backwards compatibility is imperative.)

cdent:
The recent tidying up in `runner.py` ought to make doing this a bit easier, but it is still the case that the files would need to be named an argument that either takes multiple files `-f foo.yaml,bar.yaml` or a multiple argument: `-f foo.yaml -f bar.yaml`.

I'm still not entirely convinced how doing so it better than running gabbi-run multiple times.

FND:
Since the OP didn't clearly state the primary use case (which led to an unhelpful implementation in #179), let me restate it explicitly: It would be nice to support globbing in order to target all YAML files within a test directory.

> I'm still not entirely convinced how doing so it better than running gabbi-run multiple times.

AIUI we have two options:

1. Limit `gabbi-run` to a single file at a time, thus requiring a wrapper script to handle multiple files:

    ```bash
    set -e

    for filepath in *.yaml; do
        gabbi-run … < $filepath
    done
    ```

2. Support multiple files directly within `gabbi-run`:

    ```
    $ gabbi-run … *.yaml …
    ```

    (though of course it's still the shell taking care of globbing here)

While the former is not exactly arduous, it's kind of annoying - and not actually as straightforward as it might seem (think `failfast`, error reporting, etc.). Plus it's fairly common for CLI tools to support multiple files, so the latter is not an unreasonable expectation.

That leaves the question of how to design the CLI arguments, given the aforementioned constraints WRT positional arguments and backwards compatibility. I actually think the OP's implicit suggestion is quite a good one:

    $ gabbi-run … -- *.yaml

That is, use `--` as [end-of-options delimiter](http://unix.stackexchange.com/a/147745), which is to be followed only by file paths:

    $ gabbi-run --response-handler … localhost:8000 -- *.yaml

cdent:
#180 closed this

PR #179, opened 2016-10-02
gabbi-run support for multiple input files

FND:
this addresses #104 

caveats:

* I'm no longer certain this is actually desirable (cf. @cdent's objections in the aforementioned issue)
* various implementation details seem crude (I kinda ran out of steam, so this is untypically unpolished)
* no documentation yet (this includes `help`); let's first decide whether we wanna merge this at all

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -10,7 +10,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""Implementation of a command-line runner of single gabbi files."""
+"""Implementation of a command-line runner of single Gabbi files."""
 
 import argparse
 from importlib import import_module
@@ -93,22 +93,30 @@ def run():
     host, port, prefix, force_ssl = utils.host_info_from_target(
         args.target, args.prefix)
 
-    # Initialize response handlers.
-    handler_objects = initialize_handlers(args.response_handlers)
+    response_handlers = initialize_handlers(args.response_handlers)
 
-    data = utils.load_yaml(handle=sys.stdin)
+    success = execute(sys.stdin, response_handlers, host, port, prefix,
+                      force_ssl, args.failfast)
+    sys.exit(not success)
+
+
+def execute(handle, response_handlers, host, port, prefix,  # TODO(FND): rename
+            force_ssl=False, failfast=False):
+    data = utils.load_yaml(handle)
     if force_ssl:
         if 'defaults' in data:
             data['defaults']['ssl'] = True
         else:
             data['defaults'] = {'ssl': True}
+
     loader = unittest.defaultTestLoader
     test_suite = suitemaker.test_suite_from_dict(
         loader, 'input', data, '.', host, port, None, None, prefix=prefix,
-        handlers=handler_objects)
+        handlers=response_handlers)
+
     result = ConciseTestRunner(
-        verbosity=2, failfast=args.failfast).run(test_suite)
-    sys.exit(not result.wasSuccessful())
+        verbosity=2, failfast=failfast).run(test_suite)
+    return result.wasSuccessful()
 
 
 def initialize_handlers(response_handlers):
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -88,6 +88,13 @@ def run():
         help='Custom response handler. Should be an import path of the '
              'form package.module or package.module:class.'
     )
+    parser.add_argument(
+        '-f',
+        nargs='?', default=None,
+        dest='input_files',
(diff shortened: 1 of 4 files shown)


gabbi/runner.py
```diff
@@ -88,27 +88,59 @@ def run():
...
+
+    failfast = args.failfast
+    failure = False
+    for input_file in input_files:
+        params = (response_handlers, host, port, prefix, force_ssl, failfast)
+        # XXX(FND): special-casing; use generic stream detection instead?
+        if input_file is sys.stdin:
+            success = execute(input_file, *params)
+        else:  # file path
+            with open(input_file, "r") as fh:
```
FND: Should we use `os.abspath` here?

cdent: I'm not sure it really matters one way or another, in this context. We're not accepting alien input here. We can assume that the user has some notion of what they are doing.

gabbi/runner.py
```diff
@@ -10,7 +10,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""Implementation of a command-line runner of single gabbi files."""
+"""Implementation of a command-line runner of single Gabbi files."""
```
cdent: You couldn't help yourself could you?

gabbi/runner.py
```diff
@@ -88,27 +88,59 @@ def run():
...
+        help='input files'
+    )
 
     args = parser.parse_args()
     host, port, prefix, force_ssl = utils.host_info_from_target(
         args.target, args.prefix)
 
-    # Initialize response handlers.
-    handler_objects = initialize_handlers(args.response_handlers)
+    response_handlers = initialize_handlers(args.response_handlers)
```
cdent: This variable rename seems extraneous. Though the argument remains response_handlers to preserve backward compatibility and not add noise to the system, we're in a world of content-handlers _and_ response-handlers now, thus handler_objects.

However, it doesn't really matter, just noting that I noticed.

gabbi/runner.py
```diff
@@ -88,27 +88,59 @@ def run():
...
+
+    input_files = args.input_files
+    if input_files is None:
+        input_files = [sys.stdin]
+
+    failfast = args.failfast
+    failure = False
+    for input_file in input_files:
+        params = (response_handlers, host, port, prefix, force_ssl, failfast)
```
cdent: Just out of curiosity, why are you collapsing the method arguments to params instead of preserving them?

gabbi/runner.py
```diff
@@ -88,27 +88,59 @@ def run():
...
+        # XXX(FND): special-casing; use generic stream detection instead?
```
cdent: Rather than special casing here, another option would be to assemble the arguments before checking files and then do something like:

```
if input_file is None:
    execute(...)
else:
    for input_file in input_files:
        executre(...)

File: gabbi/tests/gabbits_runner/minimal_failure.yaml [hunk 6]
```diff
@@ -0,0 +1,6 @@
+tests:
+
+- name: expected failure
+  GET: /
+
+  status: 666
```
cdent: 👍

gabbi/runner.py
FND: I'm always open to alien input; it can only get better.

gabbi/runner.py
FND: To be honest, I don't know what that really means. I'd only changed it because "handler _objects_" is far too generic to mean anything.

gabbi/runner.py
FND: `execute` is being invoked twice, so DRY (to avoid accidental inconsistencies).

gabbi/runner.py
FND: I feel like that would put too much distance between the two `execute` invocations, which also means that the processing order (in the conveyor-belt sense) becomes somewhat muddled. This way processing of CLI arguments and actual execution are nicely separated.

gabbi/runner.py
```diff
@@ -88,27 +88,59 @@ def run():
...
+        if input_file is sys.stdin:
+            success = execute(input_file, *params)
+        else:  # file path
+            with open(input_file, "r") as fh:
+                success = execute(fh, *params)
```
FND: You're right, that's a bug: Once `failure` has been set to `True`, it should be considered immutable. How about this (below):

```diff
-failure = not success
+failure = failure or not success
```

FND:
Turns out this doesn't actually address the original use case: `-f /path/to/foo -f /path/to/bar` doesn't support globbing. I'll submit a new PR after discussing options in #104.

PR #180, opened 2016-10-03
gabbi-run support for multiple input files (take two)

FND:
this addresses #104, improving upon #179 (cf. https://github.com/cdent/gabbi/issues/104#issuecomment-251101913)

no documentation yet because I wasn't sure how or where

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -61,13 +61,7 @@ def run():
 
     Output is formatted as unittest summary information.
     """
-    argv = sys.argv
-    try:  # extract file paths, separated by `--`
-        i = argv.index("--")
-        input_files = argv[i + 1:]
-        argv = argv[:i]
-    except ValueError:  # no file paths
-        input_files = [sys.stdin]
+    argv, input_files = extract_file_paths(sys.argv, default_files=[sys.stdin])
 
     parser = argparse.ArgumentParser(description='Run gabbi tests from STDIN')
     parser.add_argument(
@@ -176,6 +170,20 @@ def load_response_handlers(import_path):
         if callable(custom_handlers):
             custom_handlers = custom_handlers()
     return custom_handlers
+
+
+def extract_file_paths(argv, default_files=None):
+    """Extract command-line arguments following the `--` end-of-options
+    delimiter, if any.
+    """
+    try:  # extract file paths, separated by `--`
+        i = argv.index("--")
+        input_files = argv[i + 1:]
+        argv = argv[:i]
+    except ValueError:
+        input_files = default_files
+
+    return argv, input_files
 
 
 if __name__ == '__main__':
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -61,38 +61,12 @@ def run():
 
     Output is formatted as unittest summary information.
     """
+    parser = _make_argparser()
+
     argv, input_files = extract_file_paths(sys.argv, default_files=[sys.stdin])
 
-    parser = argparse.ArgumentParser(description='Run gabbi tests from STDIN')
-    parser.add_argument(
-        'target',
-        nargs='?', default='stub',
-        help='A fully qualified URL (with optional path as prefix) '
-             'to the primary target or a host and port, : separated. '
-             'If using an IPV6 address for the host in either form, '
-             'wrap it in \'[\' and \']\'.'
-    )
-    parser.add_argument(
-        'prefix',
-        nargs='?', default=None,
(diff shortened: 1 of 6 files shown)


cdent:
docs go in docs/source/runner.rst you'll have to experiment a bit I think to see how you want to express the options for how files work

gabbi/runner.py
```diff
@@ -10,7 +10,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""Implementation of a command-line runner of single gabbi files."""
+"""Implementation of a command-line runner of single Gabbi files."""
```
cdent: you've introduced a typo here

gabbi/runner.py
```diff
@@ -59,6 +59,13 @@ def run():
 
     Output is formatted as unittest summary information.
     """
+    argv = sys.argv
+    try:  # extract file paths, separated by `--`
+        i = argv.index("--")
+        input_files = argv[i + 1:]
+        argv = argv[:i]
+    except ValueError:  # no file paths
+        input_files = [sys.stdin]
```
cdent: I'd prefer you do this after the argparse argument declarations (or extract those to another method, call that and then argv munging).

gabbi/runner.py
```diff
@@ -89,26 +96,48 @@ def run():
...
 
-    # Initialize response handlers.
-    handler_objects = initialize_handlers(args.response_handlers)
+    response_handlers = initialize_handlers(args.response_handlers)
 
-    data = utils.load_yaml(handle=sys.stdin)
+    failfast = args.failfast
+    failure = False
+    for input_file in input_files:
+        params = (response_handlers, host, port, prefix, force_ssl, failfast)
```
cdent: still no super keen on this param globbing, when execute is called you have to look elsewhere to see what it is called with, that seems bad to me.

gabbi/runner.py
```diff
@@ -89,26 +96,48 @@ def run():
...
+
+        if not failure:  # once failed, this is considered immutable
+            failure = not success
+        if failure and failfast:
+            break
+
+    sys.exit(failure)
+
+
+def execute(handle, response_handlers, host, port, prefix,  # TODO(FND): rename
```
cdent: either accept the name or change it, do not leave for later. If you're trying to disambiguate, try `execute_tests` or `execute_test_file` or things in that vein

gabbi/runner.py
```diff
@@ -89,26 +96,48 @@ def run():
...
         if 'defaults' in data:
             data['defaults']['ssl'] = True
         else:
             data['defaults'] = {'ssl': True}
+
     loader = unittest.defaultTestLoader
     test_suite = suitemaker.test_suite_from_dict(
         loader, 'input', data, '.', host, port, None, None, prefix=prefix,
-        handlers=handler_objects)
+        handlers=response_handlers)
```
cdent: doing `s/handler_objects/response_handlers/` on the file is an unrelated change that you seem to only be doing because you don't like/understand the name. Why is the change a part of this patchset?

FND: You're right, this doesn't belong here - will fix.

gabbi/runner.py
FND: As agreed: `run_suite`

FND:
Thanks for the review. I've tried to address your concerns, rewriting history for minor changes (`run_suite`, `handler_objects`) and adding new commits for the rest.

As for docs, I'm not gonna do that tonight - so I'd prefer to create a separate issue for that, to be addressed another day.

gabbi/runner.py
FND: Assuming you're referring to "single" being obsolete, does that really count as a typo?

cdent: You appear to accidentally capitalised gabbi where it wasn't before. I assume this was a mistake of some kind?

gabbi/runner.py
```diff
@@ -89,26 +92,48 @@ def run():
...
+    args = parser.parse_args(argv[1:])
     host, port, prefix, force_ssl = utils.host_info_from_target(
         args.target, args.prefix)
 
-    # Initialize response handlers.
     handler_objects = initialize_handlers(args.response_handlers)
 
-    data = utils.load_yaml(handle=sys.stdin)
+    _run = lambda fh: run_suite(fh, handler_objects, host, port, prefix,
+                                force_ssl, failfast)
```
cdent: see my comment elsewhere about obscurity 👎

Issue #182, opened 2016-10-19
If verbose: True and response content-type is json may as well pretty print the output


cdent:
It kinda seems like if we know the content-type when being verbose about bodies, we may as well pretty print if it is json. It's not much effort to do so (there's already pretty printing happening in some test failure messages) and it is nice.

Is it too nice?

/cc @FND

PR #183, opened 2016-11-04
Verbose post bug

tomviner:
`VerboseHttp` is printing the request and response via `utils.decode_response_content`.

`decode_response_content` expects its input to require decoding, if the content_type is non-binary. However the POST body data is already decoded into a string, so decoding again fails.

```shell
E POST data with verbose true

ERROR: POST data with verbose true
	  File "/home/tom/dev/gabbi/gabbi/case.py", line 92, in wrapper
	    func(self)
	  File "/home/tom/dev/gabbi/gabbi/case.py", line 140, in test_request
	    self._run_test()
	  File "/home/tom/dev/gabbi/gabbi/case.py", line 391, in _run_test
	    redirect=test['redirects'])
	  File "/home/tom/dev/gabbi/gabbi/case.py", line 322, in _run_request
	    redirect=redirect
	  File "/home/tom/dev/gabbi/gabbi/httpclient.py", line 113, in request
	    self._print_body(headers, body)
	  File "/home/tom/dev/gabbi/gabbi/httpclient.py", line 145, in _print_body
	    utils.decode_response_content(headers, content))
	  File "/home/tom/dev/gabbi/gabbi/utils.py", line 67, in decode_response_content
	    return content.decode(charset)
	AttributeError: 'str' object has no attribute 'decode'
----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (errors=1)
```

### DIFF
--- a/gabbi/tests/test_verbose.yaml
+++ b/gabbi/tests/test_verbose.yaml
@@ -1,0 +1,9 @@
+tests:
+
+  - name: POST data with verbose true
+    verbose: true
+    POST: /
+    request_headers:
+        content-type: application/json
+    data:
+        'text'
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -360,7 +360,7 @@ class HTTPTestCase(testtools.TestCase):
                 raise exception.GabbiFormatError(
                     'malformed headers in test %s: %s' % (test['name'], exc))
 
-        if test['data'] is not '':
+        if test['data'] != '':
             body = self._test_data_to_string(
                 test['data'],
                 utils.extract_content_type(headers, default='')[0])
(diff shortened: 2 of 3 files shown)


cdent:
Thanks. I suspect this once worked properly and then stopped in the switch from response-handlers to content-handlers (which moved around some string related stuff).

cdent:
This will be in 1.27.1 which will be out in a few minutes.

tomviner:
Confirmed. Installed from PyPI. Thanks for being so responsive! :+1:

gabbi/tests/test_verbose.yaml
```diff
@@ -0,0 +1,9 @@
+tests:
```
cdent: I didn't realize when merging this that this file was not actually being used. I'll fix it while dealing with #182

tomviner: Ah thanks.

PR #186, opened 2016-11-10
Pretty print JSON bodies in verbose tests

### LINKED ISSUE #182: If verbose: True and response content-type is json may as well pretty print the output

cdent:
If a request or response body is JSON, then use the loads and dumps
methods on the JSON content handler to present it in a pretty
fashion. See the TODO in gabbi/httpclient.py for future expansion
options.

This reuses a test_verbose.yaml file added in pull request #183
that didn't seem to be doing anything.

Tests for this are in test_runner.py as it provides useful stdout
capturing infrastructure.

Fixes #182

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -19,6 +19,7 @@ import sys
 
 import urllib3
 
+from gabbi.handlers import jsonhandler
 from gabbi import utils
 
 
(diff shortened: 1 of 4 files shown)


cdent:
I'm undecided if this is suitably correct in its implementation, so cc'ing @FND and @tomviner as they both expressed interested on the related bug #182

gabbi/httpclient.py
```diff
@@ -138,11 +139,20 @@ def _print_headers(self, headers, prefix=''):
...
-                utils.extract_content_type(headers)[0]):
+        content_type = utils.extract_content_type(headers)[0]
+        if self._show_body and utils.not_binary(content_type):
+            content = utils.decode_response_content(headers, content)
+            # TODO(cdent): Using the JSONHandler here instead of
+            # just the json module to make it clear that eventually
+            # we could pretty print any printable output by using a
+            # handler's loads() and dumps(). Not doing that now
+            # because it would be pointless (no other interesting
+            # handlers) and this may approach may be entirely wrong.
```
tomviner: I think this comment makes sense. We have an XML content handler, that includes pretty print in the `dumps` method. But happy to see how this code pans out here first.

tomviner:
:+1:

gabbi/httpclient.py
cdent: Sounds like you're doing a lot of interesting gabbi stuff. If you get a chance to share more about it, that would be great.

I've started an extracted HTML content handler over at https://github.com/cdent/gabbihtml based on stuff that @FND wrote a while ago. Having a diverse collection of such things that's relatively easy to find would be a nice to have, eventually.

tomviner: I'm making enquiries about open sourcing the very small set of tools we've started putting together to work with gabbi.

Our tests for the 2 handlers we've written don't even use any interceptor currently, but the approach taken in gabbihtml looks simple enough to use.

PR #187, opened 2016-11-25
Fix pytest yield

cdent:
This branch is an effort to fix the warnings caused by yield tests as introduced by pytest 3.0. It changes over to a method suggested on the [pytest-dev list](https://mail.python.org/pipermail/pytest-dev/2016-October/003907.html). That strategy has been made to work (while maintaining ackwards compatibility with existing test suites), but it still requires having the caller import the `test_pytest` function (on which the fixture generation tooling operated) along with creating a `pytest_generate_tests` function. (see the `gabbi/tests/test_gabbits_pytest.py` file for the entrypoint).

This is more cumbersome than desired and a bit hard to document/explain but I'm not sure what to do. So, to increase visibility I'm pulling the branch into a pull request that's not ready but needs help and cc'ing the current active contributors for additional opinions (please and thank you).

/cc @FND @jasonamyers @tomviner

### DIFF
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -170,7 +170,7 @@ def py_test_generator(test_dir, host=None, port=8001, intercept=None,
             ids = []
             args = []
             for test in test_list:
-                if len(test) >=3:
+                if len(test) >= 3:
                     name, method, arg = test
                 else:
                     name, method = test
@@ -185,9 +185,9 @@ def py_test_generator(test_dir, host=None, port=8001, intercept=None,
 
 
 def test_pytest(test, result):
-    try:
+    if result:
         test(result)
-    except TypeError:
+    else:
         test()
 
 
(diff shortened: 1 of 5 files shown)


trojkat:
:+1:

gabbi/driver.py
```diff
@@ -151,15 +152,43 @@ def py_test_generator(test_dir, host=None, port=8001, intercept=None,
...
             for subtest in test:
-                yield '%s' % subtest.__class__.__name__, subtest, result
-            yield 'stop_%s' % test._tests[0].__class__.__name__, test.stop
+                test_list.append(('%s' % subtest.__class__.__name__,
+                                  subtest, result))
+            test_list.append(('stop_%s' % test._tests[0].__class__.__name__,
+                              test.stop))
+
+    if metafunc:
+        if metafunc.function == test_pytest:
```
FND: What happens if this is not the case?

gabbi/tests/test_gabbits_pytest.py
```diff
@@ -20,22 +20,19 @@
 import os
 
 from gabbi import driver
+# TODO(cdent): this test_* needs to be imported bare or things do not work
```
FND: I don't understand what this means.

gabbi/driver.py
cdent: Then other tests which may be in the same file carry on like normal tests without changes.

I tried to get rid of the conditional and just do a "modify everything!" style to see about not having `test_pytest`. The code in this block is only run if there is at least one `test_*` in the modules scope.

Issue #184, opened 2016-11-04
Data <@filename isn't relative to the YAML file


tomviner:
The [docs say](https://gabbi.readthedocs.io/en/latest/format.html#data):

> If the value is a string that begins with <@ then the rest of the string is treated as the name of a file to be loaded from the same directory as the YAML file. 

But I haven't found it works like this unless I cd into the directory containing the yaml file.

`_load_data_file` says:
```python
path = os.path.join(self.test_directory, os.path.basename(filename))
```
this does a few things:
- makes the path begin with `self.test_directory` (defaults to `.`, the current working directory)
- disguards any folders in the <@ `filename` path
- appends `filename`

This means, if I'm in `/`, and I have a test that says `data: <@cake.jpg`, I cannot run my tests as advised in the docs:
`gabbi-run -- /my/test.yaml /my/other.yaml`

`FileNotFoundError: [Errno 2] No such file or directory: './cake.jpg'`

So in our test running script, we have to start with `cd tests/`.

My preference, would be to make the path relative to the location of the yaml file. That way, I can run my tests from whatever directory.

Obviously this isn't possible when redirecting with `<` as the filename is never seen by gabbi. So I'm happy is discuss other ideas.

---

Furthermore I can't keep my test data in a child directory, unless I cd into that directory like:
```shell
cd yaml_tests/test_data
gabbi-run -- ../test_things.yaml
```

So for that reason, I'd like to allow directories like `<@test_data/filename.txt` to be included in the path.

---

Happy to write the patch, if we come up with an approach here.

cdent:
The `<@` stuff predates `gabbi-run` and was written the way you've correctly described to make sure that test suites that use the `driver.build_tests` method of finding and loading tests didn't inadvertently read data from places they shoudn't be. The most explicit way of doing that was by having only files adjacent to the yaml be readable.

As you noticed this goes a bit strange when using `gabbi-run`, especially now that there is the -f functionality.

I don't have any bright ideas on how to do this well, while still keeping build_tests constraints. I'm pinging @FND so he can comment too: he's one of the driving forces behind gabbi-run so may have some thoughts.

I'm sure we can figure something out.

FND:
Given the constraints of backwards compatibility, I guess the only reasonable way forward here is for `runner` to set a working directory - either via an environment variable (`GABBI_CWD`? though I generally dislike those) or by passing an additional parameter through `suitemaker.test_suite_from_dict`.

cdent:
[...]

PR #188, opened 2016-11-26
Allow <@ syntax to safely include subdirectories

tomviner:
Fix for one of the two issues raised in https://github.com/cdent/gabbi/issues/184

### DIFF
--- a/gabbi/tests/test_load_data_file.py
+++ b/gabbi/tests/test_load_data_file.py
@@ -1,0 +1,76 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


gabbi/tests/test_load_data_file.py
```diff
@@ -0,0 +1,76 @@
...
+"""Test loading data from files with <@.
+"""
+
+import unittest
+
+from gabbi import case
+from six.moves import mock
+
+
+@mock.patch(
```
cdent: This mock fails on python 3.3 and 3.4. I _think_ that doing a `create=True` will fix it. See [the note in the docs](https://docs.python.org/dev/library/unittest.mock.html).

cdent:
Thanks, I'll do the other half, either today or tomorrow.

tomviner:
Thanks @cdent

PR #189, opened 2016-11-27
Docs for <@file/path.txt

tomviner:
Small update to https://github.com/cdent/gabbi/pull/188

Update docs to mention filepath rather than file.

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -267,8 +267,9 @@ flexibility when doing a ``POST`` or ``PUT``. If the value is not a
 string (that is, it is a sequence or structure) it is treated as a
 data structure which is turned into a JSON string. If the value is a
 string that begins with ``<@`` then the rest of the string is treated
-as a filepath to the loaded. The path is relative to the test directory.
-If the value is an undecorated string, that's the value.
+as a filepath to be loaded. The path is relative to the test directory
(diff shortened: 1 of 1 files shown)


docs/source/format.rst
```diff
@@ -267,8 +267,8 @@ flexibility when doing a ``POST`` or ``PUT``. If the value is not a
 string (that is, it is a sequence or structure) it is treated as a
 data structure which is turned into a JSON string. If the value is a
 string that begins with ``<@`` then the rest of the string is treated
-as the name of a file to be loaded from the same directory as the YAML
-file. If the value is an undecorated string, that's the value.
+as a filepath to the loaded. The path is relative to the test directory.
```
cdent: Do you mean "be loaded" instead of "the loaded"?

It's probably worth noting here that traversing out of the local subtree is not allowed. I'd be happy with something like "The path may not include .." or something with similar meaning.

tomviner: Thanks for catching the typo. I've clarified it.

Just to note, as you can see from the tests, technically you are allowed the string `..`, but you may not end up with a path, once canonicalised, where the relative path from test_directory to the one given starts with `..`.

cdent: Yeah, what you've done is better, thanks.

cdent:
thanks

PR #191, opened 2016-11-27
Use the dirname of each runner testfile as the test_dir

### LINKED ISSUE #184: Data <@filename isn't relative to the YAML file

cdent:
When gabbi-run is passed one or more files, set the working
directory for the associated test suite as the directory of
each test file. Doing so ensures that `<@file` syntax in the
YAML has a fixed point from which to find the data files.

If data comes from STDIN then the test directory is '.' (as
it has always been).

Fixes #184 

/cc @tomviner @FND

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -14,6 +14,7 @@
 
 import argparse
 from importlib import import_module
+import os
 import sys
 import unittest
 
@@ -84,8 +85,9 @@ def run():
     else:
         for input_file in input_files:
             with open(input_file, 'r') as fh:
+                data_dir = os.path.dirname(input_file)
                 success = run_suite(fh, handler_objects, host, port,
-                                    prefix, force_ssl, failfast)
+                                    prefix, force_ssl, failfast, data_dir)
             if not failure:  # once failed, this is considered immutable
                 failure = not success
             if failure and failfast:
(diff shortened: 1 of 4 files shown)


FND:
lgtm 👍

Issue #185, opened 2016-11-04
Should be able to set verbose from command line


tomviner:
The `verbose` flag is most helpful when debugging. If I've got a run of tests in a yaml file, each dependant on the one before, when I get an error, I'd like to be able to turn on verbose mode to see what's happening from the first test onward. Ideally I could change one thing in one place, like adding `--verbose=headers` or similar.

Would you be open to considering this?

cdent:
Yeah, that would be reasonable, and could be done in the same way that `force_ssl` is done in `gabbi/runner.py`.

PR #192, opened 2016-11-27
Add a -v or --verbose flag to gabbi-run

cdent:
A value of 'all', 'headers' or 'body' will have the expected
effect on all the tests being run in the current testing session.

This is done by manipulating the 'defaults' of each test suite.

Fixes: #185 

/cc @FND @tomviner

### DIFF
--- a/docs/source/runner.rst
+++ b/docs/source/runner.rst
@@ -50,3 +50,6 @@ YAML will default to ``ssl: True``.
 
 If a ``-x`` or ``--failfast`` argument is provided then ``gabbi-run`` will
 exit after the first test failure.
+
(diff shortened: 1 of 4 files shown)


FND:
Nice, thanks for this!

Only one minor concern: It seems unusual to parameterize `-v`, so it _might_ be nice if `gabbi-run --verbose` behaved like `gabbi-run --verbose=all`? But IIRC  argparse doesn't allow for that kind of flexibility.

cdent:
> But IIRC argparse doesn't allow for that kind of flexibility.

Yeah, I looked into that and couldn't figure it out so gave up with hope for future generations.

PR #193, opened 2016-12-01
gabbi-run: Fix usage of STDIN

sileht:
Recent change have break gabbi-run with STDIN.
"verbosity" is passed the the "datadir" instead of "."

This fixes that.

Also, I don't get why the test suite doesn't fail.

I have discover the issue here: http://logs.openstack.org/09/405109/1/check/gate-telemetry-dsvm-integration-ceilometer-ubuntu-xenial/b3544ba/console.html#_2016-12-01_06_44_54_763450

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -84,7 +84,8 @@ def run():
 
     if not input_files:
         success = run_suite(sys.stdin, handler_objects, host, port,
-                            prefix, force_ssl, failfast, verbosity)
+                            prefix, force_ssl, failfast,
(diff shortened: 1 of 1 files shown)


cdent:
> Also, I don't get why the test suite doesn't fail.

Yeah, that's weird. I'll merge this, release a new version, and then see about making a test that gets things more proper.

Thanks for discovering and fixing this.

cdent:
Ah, looks like there are no stdin-based tests that do data loading, that should be easy to fix.

Issue #149, opened 2016-06-10
Q: Persist (across > 1 tests) value to variable based on JSON response?


josdotso:
Example algorithm to be expressed in YAML:
- Create an object given an object name.
- Store in "$FOO" (or similar) the UUID given to object per JSON response.
- Do unrelated tests.
- Perform a GET using "$FOO" (the UUID)

Thanks as always!

josdotso:
Looks like a duplicate of #78.  I'll let @cdent decide if this should be closed though.

cdent:
One of the main issues with this general idea (also discussed on #78) is how to come up with a reasonable scenario to reference back multiple steps. Your idea takes a somewhat different spin in that you want to be able to say: assign this result to a variable.

I continue to worry that that level of flexibility will lead to unreadable tests and readable tests is one of the main goals of gabbi. However, if you have some ideas on how the syntax might work, please stick them on here so we can hash through it. Enough people have asked for it that it is probably worth considering. For some reason I find your idea a bit more appealing. It feels like the syntax could be less fugly if done right.

samuelfekete:
We need this as well, for an S3 multi-part upload.

I think a nice interface would be to have something like: `$HISTORY['Test Name'].$RESPONSE` (or `$LOCATION` etc). This would make it clear where the result is from, and it won't break the existing usage, which would still be the previous request.

This can be implemented by going up the chain of `prior.prior.prior...` until a test with the given name is found, then using that in the replacers instead of `prior`.

cdent:
I continue to be reluctant about this for the reasons stated in #78.

Which is not to say "no, this is never going to happen", just that it's something where we should make sure there are no other options.

The interface suggested by @samuelfekete is certainly better than anything we've thought of before (especially as it is safe in the face of the number of tests in a suite changing). @FND and @jasonamyers, you want to weigh in?

samuelfekete:
I can think of workarounds, but they would be much uglier. Like, have an endpoint I can call that acts as a dictionary.

samuelfekete:
Happy to contribute a with a pull request if this approach is acceptable.

jasonamyers:
I really like the idea of $HISTORY. That could be very useful for some test cases I've been avoiding.  I'm not a fan of the multiple `prior` approach as I find it to be very unreadable.  Storing of the responses by test name is nice.  I'm curious how we might need to handle parallel testing. Would we need to account for something like a `depends_on` so that we can ensure they are run in the right order?

[...]

PR #194, opened 2016-12-20
Add $HISTORY variable

samuelfekete:
Add `$HISTORY` magic variable that can be placed before other magic variables to make them refer to any previous test, e.g.  `$HISTORY['test name'].$LOCATION`.
https://github.com/cdent/gabbi/issues/149

Also added `$URL`, because `$HISTORY...$LAST_URL` would be too confusing.

### DIFF
--- a/gabbi/tests/gabbits_intercept/backref.yaml
+++ b/gabbi/tests/gabbits_intercept/backref.yaml
@@ -73,3 +73,13 @@ tests:
 - name: backref json fail end
   xfail: true
   url: $RESPONSE['url']
+
+- name: get a historical response
+  GET: $HISTORY['post some json'].$RESPONSE['a']
+  response_headers:
+      x-gabbi-url: $SCHEME://$NETLOC/1
+
+- name: get a historical response via jsonpath
+  GET: $HISTORY['post some json'].$RESPONSE['$.b']
+  response_headers:
+      x-gabbi-url: $SCHEME://$NETLOC/2
--- a/gabbi/tests/gabbits_intercept/backref.yaml
+++ b/gabbi/tests/gabbits_intercept/backref.yaml
@@ -75,11 +75,11 @@ tests:
   url: $RESPONSE['url']
 
 - name: get a historical response
-  GET: $HISTORY['post some json'].$RESPONSE['a']
+  GET: /$HISTORY['post some json'].$RESPONSE['a']
   response_headers:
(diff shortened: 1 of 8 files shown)


samuelfekete:
I think this would break the gabbihtml handler (https://github.com/cdent/gabbihtml/blob/master/gabbihtml/handler.py#L97), which uses `gen_replacer()` and not `replacer()` and it assumes that only the last response will be handled.

cdent:
> I think this would break the gabbihtml handler

That's not a problem, it's not released or really properly done yet. I'll go back through history and try to remind myself why gen_replacer (probably an early optimization explicitly for the html handler) was created just to make sure some edge case isn't getting messed up.

I'll need to give this a closer look later in the day, but a superficial read through it looks sane and workable. Nicely done.

We'll of course need to figure out what's up with the tests, and think about how to document this well.

gabbi/tests/gabbits_intercept/backref.yaml
```diff
@@ -73,3 +73,13 @@ tests:
 - name: backref json fail end
   xfail: true
   url: $RESPONSE['url']
+
+- name: get a historical response
+  GET: $HISTORY['post some json'].$RESPONSE['a']
```
cdent: In py35 `urlunsplit` has slightly different behavior so in order for this test to work properly (and to have a correct url) this needs to be:

    GET: /$HISTORY['post some json'].$RESPONSE['a']

gabbi/tests/gabbits_intercept/backref.yaml
```diff
@@ -73,3 +73,13 @@ tests:
...
+  response_headers:
+      x-gabbi-url: $SCHEME://$NETLOC/1
+
+- name: get a historical response via jsonpath
+  GET: $HISTORY['post some json'].$RESPONSE['$.b']
```
cdent: and here

samuelfekete:
@cdent, thanks for the quick response. I am not going away. Will do another commit soon to fix the failing tests.

jd:
👍 for the feature. I am not competent enough to validate the code. but this is the way I envisioned the functionality for a long time.

cdent:
I'll do some docs and release notes for this next week, unless someone gets there first. Thanks very much @samuelfekete

samuelfekete:
Thanks @cdent

cdent:
This has been released as `1.30.0`: https://pypi.python.org/pypi/gabbi/1.30.0

PR #198, opened 2017-01-06
add application/xml.* as non-binary

tomviner:
We're testing an S3 compatible API, and verbose=all doesn't show our errors, as gabbi thinks `application/xml` is a binary mime type.

I've used `startswith` so that a mime type like `application/xml-dtd` (Document Type Definition) also gets tagged non-binary.

### DIFF
--- a/gabbi/tests/test_utils.py
+++ b/gabbi/tests/test_utils.py
@@ -32,6 +32,7 @@ class BinaryTypesTest(unittest.TestCase):
         'application/javascript',
         'application/json',
         'application/json-home',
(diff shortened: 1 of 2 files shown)


FND:
Looks good - thank you!

cdent:
https://pypi.python.org/pypi/gabbi/1.30.1

tomviner:
Thanks, and a release, that's great! :smiley:

cdent:
You keep sending easy to comprehend reasonable fixes and improvements, we'll keep releasing them.

PR #200, opened 2017-01-13
Handle exceptions where starting fixtures

### LINKED ISSUE #199: exceptions that happen in start_fixture() are swallowed during the test discovery process
The discovery will fail and end up producing a seemingly non-related error like `math domain error` and the real exception is nowhere to be found.

This needs to be narrowed to a minimal test case and then figured out.

What seems likely is that something about the test suites is being processed is not following unittest/testtools rules of failure.

cdent:
Some test runners will swallow exceptions and tracebacks that happen
during the start_fixture method of a Fixture. This can lead to a
great deal of confusion when creating new fixtures. This change
ensures that the traceback is made available by forcing an error to
be reported on the first test in the test suite using the fixture.
Additional detail can be found in the comment associated with the
change.

Fixes #199

### DIFF
--- a/gabbi/suite.py
+++ b/gabbi/suite.py
@@ -16,6 +16,7 @@ This suite has two features: the contained tests are ordered and there
 are suite-level fixtures that operate as context managers.
 """
 
+import sys
 import unittest
 
 from wsgi_intercept import interceptor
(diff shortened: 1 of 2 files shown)


cdent:
I'm struggling (because of other mental commitments and general stupidness) to come up with a reasonable way to add tests for this. If anyone has ideas please feel free to let me know.

PR #201, opened 2017-02-02
Adjust runner to track test suite failures

### LINKED ISSUE #181: gabbi-run does not expose failures' originating suite
To reproduce:

```
$ python -m gabbi.runner http://example.org -- gabbi/tests/gabbits_runner/*.yaml
```

(which corresponds to `gabbi-run … -- /path/to/*.yaml`)

This results in three failures from three separate suites:

```
... ✗ expected failure

FAIL: expected failure
    …
    AssertionError: '301' not found in ['666'], response:
    …
----------------------------------------------------------------------
Ran 1 test in 0.183s

FAILED (failures=1)
... ✗ expected success

FAIL: expected success
    …
    AssertionError: '301' not found in ['200'], response:
    …
----------------------------------------------------------------------
Ran 1 test in 0.065s

FAILED (failures=1)
... ✗ expected success

FAIL: expected success
    …
    AssertionError: '301' not found in ['200'], response:
    …
----------------------------------------------------------------------
Ran 1 test in 0.060s

FAILED (failures=1)
```

Other than the respective test's `name`, there's barely any indication where a failure originated.

Somewhat relatedly, there's also no summary across all suites - so a casual observer might only spot the last report, erroneously concluding there was only a single failure (or worse: if the last suite did _not_ fail, assume everything's fine).

cdent:
Change the reporter used by gabbi-run so that it provides a more
complete test name in its output, using the id() of the test. This
will include the name of the yaml file.

Also track failures per input file and provide a summary of which
files contained failures (if any) at the end.

Both of these changes were done in the simplest way possible so
there is probably considerably room for improvement now that the
concept has been proven.

Fixes #181

### DIFF
--- a/gabbi/reporter.py
+++ b/gabbi/reporter.py
@@ -80,7 +80,8 @@ class ConciseTestResult(TextTestResult):
             self.stream.writeln('\t[unexpected success]')
 
     def getDescription(self, test):
-        name = test.test_data['name']
+        # Chop the test method (test_request) off the test.id().
+        name = test.id().rsplit('.', 1)[0]
         desc = test.test_data.get('desc', None)
         return ': '.join((name, desc)) if desc else name
 
--- a/gabbi/reporter.py
+++ b/gabbi/reporter.py
@@ -80,7 +80,7 @@ class ConciseTestResult(TextTestResult):
             self.stream.writeln('\t[unexpected success]')
 
     def getDescription(self, test):
-        # Chop the test method (test_request) off the test.id().
+        # Chop the test method ('test_request') off the test.id().
         name = test.id().rsplit('.', 1)[0]
         desc = test.test_data.get('desc', None)
         return ': '.join((name, desc)) if desc else name
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -12,6 +12,8 @@
 # under the License.
 """Implementation of a command-line runner for gabbi files (AKA suites)."""
 
+from __future__ import print_function
+
 import argparse
 from importlib import import_module
 import os
@@ -81,6 +83,8 @@ def run():
     verbosity = args.verbosity
     failfast = args.failfast
     failure = False
+    # Keep track of file names that have failures.
+    failures = []
 
     if not input_files:
         success = run_suite(sys.stdin, handler_objects, host, port,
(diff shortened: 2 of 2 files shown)


cdent:
@FND it would be good if you could look at this and see how to make it more aligned with your expectation for output and naming. But this should give you the framing to make some thing nice.

gabbi/reporter.py
```diff
@@ -80,7 +80,8 @@ def addUnexpectedSuccess(self, test):
             self.stream.writeln('\t[unexpected success]')
 
     def getDescription(self, test):
-        name = test.test_data['name']
+        # Chop the test method (test_request) off the test.id().
```
FND: Should "test_request" be quoted here? I'm genuinely unsure what it means.

gabbi/reporter.py
```diff
@@ -80,7 +80,8 @@ def addUnexpectedSuccess(self, test):
...
+        name = test.id().rsplit('.', 1)[0]
```
FND: This seems like breaking encapsulation, i.e. ideally that functionality would be exposed by `test`'s class (not sure what that'd be, exactly)?

gabbi/runner.py
```diff
@@ -89,22 +93,28 @@ def run():
         failure = not success
     else:
         for input_file in input_files:
+            name = os.path.splitext(os.path.basename(input_file))[0]
```
FND: Similar to my encapsulation-related comment above, it'd be nice if this logic could be moved somewhere else - though I'm aware that's perhaps a little trickier in this case.

gabbi/runner.py
```diff
@@ -89,22 +93,28 @@ def run():
...
             with open(input_file, 'r') as fh:
                 data_dir = os.path.dirname(input_file)
                 success = run_suite(fh, handler_objects, host, port,
                                     prefix, force_ssl, failfast,
                                     data_dir=data_dir,
-                                    verbosity=verbosity)
+                                    verbosity=verbosity, name=name)
+            if not success:
```
FND:
```python
if not success:
    …
if not failure:
    …
```
strikes me as odd - was that intentional?

gabbi/runner.py
```diff
@@ -120,8 +130,8 @@ def run_suite(handle, handler_objects, host, port, prefix, force_ssl=False,
 
     loader = unittest.defaultTestLoader
     test_suite = suitemaker.test_suite_from_dict(
-        loader, 'input', data, data_dir, host, port, None, None, prefix=prefix,
-        handlers=handler_objects)
+        loader, name, data, data_dir, host, port, None, None, prefix=prefix,
+        handlers=handler_objects, test_loader_name='gabbi-runner')
```
FND: As a naive user, that "gabbi-runner." prefix strikes me as pointless noise - can you explain the significance from your perspective?

gabbi/reporter.py
cdent: probably

gabbi/reporter.py
cdent: It's the reporter that is weird here, not test case, so I wanted to put the weirdness here.

gabbi/runner.py
cdent: It was intentional pattern replication because sadly success and failure aren't meaning the same things. success is a value for right now, failure is a value for since we started.

gabbi/runner.py
cdent: without that the test name is `gabbi.suitemaker.some.stuff`. this makes it `gabbi-runner.some.stuff`.

cdent:
I'll make some tidies based on your suggestions. Thank you.

PR #204, opened 2017-02-16
Unicode content from disk must become bytes

cdent:
While trying to figure out some tests for the fix in #203 it became
clear that the handling of encoded-on-disk-but-textual content
when using <@ to load content was not sufficiently robust in both
python 2 and 3.

We want to load content to become unicode so that it is used
internally correctly, for example in the replace_template handling
used for substitutions, but when an HTTP request is made the content
must become whatever the versioned equivalent of bytes is.

Therefore the code has been changed to load non_binary content and
decode it to six.text_type, do replace_template on it, and then turn
the body, if it is not already bytes, into bytes.

This is verified by adjusting data.json to have some utf8 in it and
adding and loading a utf8.txt example file.


gabbi/tests/gabbits_intercept/utf8.txt
```diff
@@ -0,0 +1,222 @@
+<h1>Unicode Demo</h1>
+
+<p>Taken from <a
+href="http://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-demo.txt">http://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-demo.txt</a></p>
+
+<pre>
+
+$ENVIRON['HOME']
```
tomviner: Did you mean to include this? Is it a new feature, replacements in data files?

cdent: It's always been a feature (but probably not well documented) and until you started using the method for response checking, was the main reason the data was being decoded from bytes to text: so the regular expressions could work. If the replace_template method wasn't being called there would be a nearly direct path to the data that was being read used as a raw body.

However, following the rule of thumb that we should be bytes at the io boundaries and text internally, the change to text and then back to bytes, is a good thing.

Make sense? And if so, do you concur?

cdent: I felt obliged to check: it is documented in [substitution](http://gabbi.readthedocs.io/en/latest/format.html#substitution) but glancingly. `data` is included in the list of fields where the variables may be used. That could probably be improved.

tomviner: Yes, that makes sense. Compare as text when we have text, and binary if not.

Ah right, replacement in data files is indeed a documented feature.

PR #205, opened 2017-02-22
Avoid duplicating prefix when re-using urls

### LINKED ISSUE #165: url concatenation with a prefix doesn't account for the prefix in the right hand side
If, for example, a gabbi-run is done with a prefix in the url:

```
gabbi-run http://example.com/prefix
```

against a service which produces output that has links which are absolute urls:

``` json
{"links": [{"rel": "self", "href": "/prefix/somewhere"}]}
```

following that link:

``` yaml
# assume a prior request
tests:
  name: get self
  GET: $RESPONSE['$.links[?rel = "self'].href']
```

will result in a 404 because the generated url will be GET `http://example.com/prefix/prefix/somewhere`.

This can be worked around in existing tests by using the prefix in the tests, not in the host info given to gabbi.

This is not, however, a good fix because the point of the prefix is to allow testing of services that might be mounted in different "sub directories".

cdent:
If a reponse includes an absolute link, re-using that link as the
url in a subsequent test when a prefix is being used in the tests
can result in a 404 because the prefix will be duplicated.

This change ensures that the prefix will not be duplicated if the
URL already begins with the prefix.

Fixes #165


cdent:
It's not clear to me if this is the right solution. It breaks when a URL such as `/prefix/prefix/foo/bar` is actually legit in the system. Is that a problem?

/cc @FND

cdent:
Note that `urlparse.urljoin` is not workable for this stuff because it only joins if the second argument doesn't begin with `/` and the first argument ends with `/`, which is just not how the concept of "prefix" works.

PR #206, opened 2017-03-03
Coerce JSON types into correct values for later $RESPONSE replacements

### LINKED ISSUE #147: substitutions result in string which can be a problem for JSON

justanotherdot:
Resolves #147, having $RESPONSE replacements that contained integer or decimal values would wind up as quoted strings after substitution, e.g. `{"id": 825}` would later become `{"id": "825"}`. 

This passes `tox -epep8`, but running `tox -epy27` seems to have the first few tests fail, unfortunately. This patch works for my use case, however; I could not POST data to particular endpoints of an API as the number values were wrapped in quotes (and the API was doing type checks on the values). 

This may not be an ideal (or eloquent) solution, but I've also tried to keep performance in mind in lieu of large JSON response bodies, namely that I expect the exception cases to be more common than exceptional, so I've added some additional checking to see if it's really worth parsing particular values as strings (line no. 359) and or if it even looks like a number in the first place (line no. 376). That said, if this performance hit is not an issue, it certainly is a lot more readable without the checks. 

I also chose to do two try/excepts instead of simply using `float`. Firstly we try parsing for int and then for float as I prefer the resulting JSON to be correct i.e. I would rather not an `id` field that was initially an int be cast into a float. For example, consider we get a response of `{"id": 825}` and we simply had one try/except that used float. The value would parse, but the resulting JSON (from `json.dumps`) would be `{"id": 825.0}`. This pragmatically doesn't matter as I'm sure most endpoints will accept a decimal value with an appended `.0` to be valid as an integer, but I felt the semantics would be a surprise to other users of the lib and it's still possible that certain APIs might have an issue with.

And thanks for all the effort you've put into the lib!

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -201,6 +201,10 @@ class HTTPTestCase(testtools.TestCase):
         """Replace an indicator in a message with the environment value."""
         value = re.sub(self._replacer_regex('ENVIRON'),
                        self._environ_replacer, message)
+        if value == "False":
+            return False
+        if value == "True":
+            return True
         return value
 
     @staticmethod
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -20,10 +20,10 @@ made using urllib3. Assertions are made against the reponse.
 from collections import OrderedDict
 import copy
 import functools
+import json
 import os
 import re
 import sys
-import json
 from testtools import testcase
 import time
 from unittest import result
@@ -356,9 +356,13 @@ class HTTPTestCase(testtools.TestCase):
         string = re.sub(self._replacer_regex('RESPONSE'),
                         self._response_replacer, message)
         try:
-            return self._json_loads_coerce_types(string)
+            if string[0] == '{':
+                # Looks like a JSON object.
+                return self._json_loads_coerce_types(string)
+            else:
+                return self.json_loads(string)
         except (ValueError, AttributeError):
-            # `string` was not a valid JSON representation.
+            # `string` was neither a JSON object or a number.
             return string
 
     @staticmethod
@@ -366,12 +370,19 @@ class HTTPTestCase(testtools.TestCase):
         """Ensure that all numerical JSON values are properly parsed"""
         json_rep = json.loads(string)
         for key, val in json_rep.items():
-            if isinstance(val, basestring):
+            # Since we expect the exception to happen often,
+            # test to see if it's worth even attempting to parse
+            # i.e. it looks like a number.
+            if isinstance(val, six.string_types) and val[:1].isdigit():
                 try:
-                    num_val = float(val)
-                    json_rep[key] = num_val
+                    int_val = int(val)
+                    json_rep[key] = int_val
                 except ValueError:
-                    # Value is not a number and does not need coercing.
+                    pass
+                try:
+                    float_val = float(val)
+                    json_rep[key] = float_val
+                except ValueError:
                     continue
         return json.dumps(json_rep)
 
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -495,6 +495,14 @@ class HTTPTestCase(testtools.TestCase):
                 else:
                     return info
         else:
+            if isinstance(data, dict):
+                data_new = {}
+                for k, v in data.items():
+                    if isinstance(v, six.string_types) and all((r not in v for r in REPLACERS)):
+                        data_new[k] = str('\"{}\"'.format(v))
+                    else:
+                        data_new[k] = v
+                data = data_new
             content_handler_cls = self.get_content_handler(content_type)
             if content_handler_cls:
                 data = content_handler_cls.dumps(data)
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -155,7 +155,8 @@ class HTTPTestCase(testtools.TestCase):
                                                    content_handler_cls)
             return message
         if isinstance(message, list):
-            return [self.replace_template(val, content_handler_cls) for val in message]
+            cls = content_handler_cls
+            return [self.replace_template(v, cls) for v in message]
 
         for replacer in REPLACERS:
             template = '$%s' % replacer
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -20,7 +20,6 @@ made using urllib3. Assertions are made against the reponse.
 from collections import OrderedDict
 import copy
 import functools
-import json
 import os
 import re
 import sys
@@ -148,7 +147,7 @@ class HTTPTestCase(testtools.TestCase):
                 return handler
         return None
 
-    def replace_template(self, message):
+    def replace_template(self, message, content_handler_cls=None):
         """Replace magic strings in message."""
         if isinstance(message, dict):
             for k in message:
@@ -166,6 +165,13 @@ class HTTPTestCase(testtools.TestCase):
                         raise AssertionError(
                             'unable to replace %s in %s, data unavailable: %s'
                             % (template, message, exc))
+                    if content_handler_cls:
+                        try:
+                            message = content_handler_cls.coerce(message)
+                        except (KeyError, AttributeError, ValueError) as exc:
+                            raise AssertionError(
+                                'unable to coerce types in %s: %s'
+                                % (message, exc))
             except TypeError:
                 # Message is not a string
                 pass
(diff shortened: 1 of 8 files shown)


justanotherdot:
Apologies, I think I pushed an earlier revision. I'll be pushing a more relevant commit in a second.

justanotherdot:
Actually, just read more into #161 and why you abandoned that change. I take it that you want an approach that isn't just assuming everything is a number but can work with custom handlers?

cdent:
This is interesting but it's not really as complete as what #161 would have been, had it been completed. In that one, all the [substitutions](http://gabbi.readthedocs.io/en/latest/format.html#substitution) are handled, not just $RESPONSE. The reason I abandoned it is the code is old enough that brining it up to date with the changes from response handlers to content handlers (at version 1.26.0) as well as the adjustments made to get $HISTORY (v 1.30.0) were fiddly enough to make me want to start again from scratch. The original concept there (making changes in `replace_template`) is likely still sound: It tries to do the casting to float or int once it has already extracted an atom, not while processing an entire blog of data.

The use case that you describe (trying to POST to something that is enforcing types on the server side with jsonschema) is one of the ones that #161 is trying to solve, but not the only one: Numeric values in `$ENVIRON` is another main one.

If the above makes any sense, and you want to try again from the replace_template angle, I'd be happy to help bring that to some form of merge-able. If not, it's on my list of things to do but I'm currently in a bit of a crunch on other things.

Some other comments within the code too.

gabbi/case.py
```diff
@@ -352,8 +353,38 @@ def _simple_replacer_regex(key):
...
-        return re.sub(self._replacer_regex('RESPONSE'),
-                      self._response_replacer, message)
+        string = re.sub(self._replacer_regex('RESPONSE'),
+                        self._response_replacer, message)
+        try:
+            if string[0] == '{':
+                # Looks like a JSON object.
+                return self._json_loads_coerce_types(string)
+            else:
+                return self.json_loads(string)
```
cdent: I'm guessing you probably wanted `json.loads` here, not `self.json_loads` (which doesn't exist).

Also, for sake of tidiness we've made the habit of not having JSON processing in this file and instead do most JSON related things in `gabbi.handlers.jsonhandler`, even if it is just utility functions or classmethods that are json-ish, not specifically related to content handlers.

cdent:
Looking over my own code again, and realizing I've not had enough coffee, I'm not entirely sure that my first comment above is 100% correct, but the gist is: To get things to work across all the substitutions, not just $RESPONSE, `replace_template` is a good starting point. Ideally whatever the solution it would come with a lot of tests that try to explore the corner cases.

Thanks a lot for looking into this. It's a tricky thing that's made more tricky by the way in which some of the functionality grew organically so getting some clarity can be hard. I'm sure we can work something out with a bit of iteration.

With regard to you comments about performance: The string and digit tests make sense (whatever the solution is and wherever we put it in the code) but I wouldn't stress too much about performance until we are sure we have a fully working solution. If it proves to be slow we can tune from there.

gabbi/case.py
justanotherdot: Ah yup; funny I didn't run into that while building my tests.

justanotherdot:
I'm more than happy to try to take another stab as it's helpful to get your take on the situation and my changes. Admittedly what I'm doing here is a bit of a hack just to coerce $HISTORY and $RESPONSE (at least those seem to work for me) but it's a sound point to try to coerce all possible replacements. 

I'll start fresh from a master version of `case.py` and go from there and commit here instead of submitting a new PR, does that sound alright? Otherwise, happy to make a new PR.

justanotherdot:
I've provided a `coerce` method inside of the base ContentHandler which just acts like `identity`. The JSONHandler then is free to decide how it should coerce values. All values are coerced after substitutions. Thankfully `json.loads` knows what to do given an int, a float, on object, an array, etc. The downside is that the JSON always has to be quoted, but I think that's normal in YAML, actually, e.g. `key: '{"key": "val"}'`.

Also, this doesn't ever use the base ContentHandler, which might actually cause problems being `identity`. It could, instead, simply `pass` instead of returning a value.

justanotherdot:
Actually, this seems to have problems when the data object has multiple keys, so I'll look into that, too.

justanotherdot:
Seems like ENVIRON is still having problems being 'coerced' so I'll have to mull over it a bit more; so long as this is a step in the right direction, of course.

justanotherdot:
Tested on my actual use cases today and there were a few problems; in particular I was overwriting the same object I was iterating over (foolishly), as well as not checking for TypeError in addition to ValueError in JSONHandler as well as some small differences. 

I'm not at the computer I was at with the test cases, however, so I'll need to double check this, but I think I have a reasonable approach that may or may not be passing the test cases again. I'll try to look into why.

Let me know if any of this looks fishy or not the way you'd like to go forward.

cdent:
Conceptually this seems sound [1] but it would be great if you could build some tests into the patch. The simple wsgi app that is down in `gabbi/tests` makes it possible to round trip data through a series of tests and thus see how the data gets transformed.

Having a complete set of tests for all the substitutions especially `$ENVIRON` would help move things along.

If I get some time a bit later in the day I'll check this out and do some of my own experimenting and see if I can come up with some tests and identify any issues.

[1] Initially I wasn't quite sure because I was worried about substitutions other than within `data` but it's really only there where it actually matters.

justanotherdot:
Sure, would be more than happy to chuck some tests in. I need to fix a few things as well. I think the test that's failing for me is because replace template in the JSON path assertions is not taking any content handler so the vales aren't getting coerced, thus it's comparing a string to an int. 

Ill be able to push a few of the changes but probably not the tests. Is it best I close this until I've got those ready?

cdent:
> Ill be able to push a few of the changes but probably not the tests. Is it best I close this until I've got those ready?

No, no reason to close it. When it is ready to go just say so. It's usually only me (of three people) that does the merging. If you want to be especially safe you could commit one failing test.

justanotherdot:
Tests to come; maybe on the weekend. Please take a look at the JSONHandler changes in particular, though. What I had to do in `action` could probably be nicer, but this made sense in the context of why the tests were failing.

cdent:
I've played around with this a bit, making a yaml file that exercises it. If you put https://gist.github.com/cdent/ec00b89b36fb7cf937a935fe041e200d in `gabbi/tests/gabbits_intercept/coerce.yaml` and then add a `ONE=1` in the `setenv` section near the top of `tox.ini` you can try it out with `tox -epy27 -- coerce` and see if it is meaningful to you. Note the two tests at the end that are marked `xfail: true`. This is because they do in fact fail. The first because lists are not being supported in coerce, and the second because nested data structures are not supported. I think some kind of recursive handing is going to be required, 

The changes in `action` seems okay. It's aligned with the fundamental assumption we're making here: If it looks like a number make it a number.

justanotherdot:
I wrote up a test yesterday that found similar issues and I've written fixes for it (nested structures and handling for lists.) I also noticed an odd edge case that sometimes the replacements would occasionally sub in Python datatypes and stringify it (e.g. `"{u'key': u'val'}"`) so I added support for that as well. 

With these changes, however, your first test of `coerce.yaml` points out something interesting. If we recursively coerce a structure, then value types that are stringy but have a datatype inherent in them are going to be parsed into their valid types: in particular `one_string` is going to become an integer and not a string.

Another thing I came up against was that stringy datatypes passed as environment variables might be too tricky at this stage to properly parse. Consider the JSON string `{"inner_key": "val"`; after substitution as a value to a key, this will cause problems for parsers as the conflicting double quotes: `{"key": "{"inner_key": "val"}"}`. At the moment I'm simply assuming that environment variables are purely for primitive types. Theoretically, this is something that could be implemented later instead of in this PR. 

I'll push my changes but I'll need to think some more about how to approach fixing some of the quirks.

justanotherdot:
Re: the 2nd paragraph above, the particular problem seems (at least to me) that when we have literals inside the yaml that are quoted strings, we want to somehow tag them to mean "please don't coerce this, this was meant as a string". However, since all replacements tend to wrap the values replaced in quotation marks, we can't make that assumption (unless we have a sensible tagging system that works in both cases, which might be possible.)

justanotherdot:
If you remove the comments on the test `object with list` in `coerce.yaml` the tests will fail for `tox -epy27 -- coerce`, and it seems to be with the expectations for what types are present in the function `assert_in_or_print_output`, namely that (in that case) the `expected` param will be a dict and the `iterable` _I think_ will be a string (given that it's the test output). I haven't tested enough to verify this but I think I can snub up a fix. I've included a few of my own tests and some are commented out which need some more changes since I was testing them before with gabbi-run manually for diagnostic purposes.

gabbi/case.py
```diff
@@ -482,13 +495,23 @@ def _test_data_to_string(self, data, content_type):
...
                     return info
         else:
-            dumper_class = self.get_content_handler(content_type)
-            if dumper_class:
-                data = dumper_class.dumps(data)
+            if isinstance(data, dict):
+                data_new = {}
+                for k, v in data.items():
+                    if isinstance(v, six.string_types) \
+                       and all((r not in v for r in REPLACERS)):
```
justanotherdot: This might need to have a similar `$` prefix as is the case in `_replace_template`, just in case someone has a RESPONSE or ENVIRON word in the text.

justanotherdot:
tl;dr of all the updates I've done today:

* Handle the issues that were present in in the handlers for response_strings which now can handle dicts and lists if so given.
* Patch in some remaining tests, including doing some minor checking for unicode support.
* Make sure that the `response_json_paths` action is doing a field-wise comparison instead of string-wise as values sometime were out of order but equality was otherwise correct. 

There still seem to be Unicode issues in the failing gnocchi tests on the CI build. I'll have to think about what's going on there, but if you have any thoughts, let me know. It's a bit harder to diagnose as I'm not completely sure how to run the gnocchi tests locally.

justanotherdot:
No worries, Hope you feel better soon!

As a note, looking at the gnocchi failing tests I feel as though the primary problem is where Unicode code points are being injected instead of their symbols (e.g. ∧ vs. \u2227). Most of this is showing up in the URLs (I think which are being generated from replacements) and a funky issue where it looks like it's trying to access some data structure with a Unicode key, but I could be totally wrong here, as well.

cdent:
Just checking in to say: I haven't forgotten about this but it has been a tempestuous time, still on my radar.

justanotherdot:
#214 resolved this PR.

PR #207, opened 2017-03-11
Allow substitutions in count and delay

### LINKED ISSUE #150: Poll and count don't use substitutions

cdent:
Since we know that count must always be an int and delay can be
either an int or float, the yaml data is passed through
replace_template and then cast to int or float explicitly.

This is different previous solution which tries to rely on the
casting happening in replace_template. With the new coerce style of
numbers-in-data handling being developed in #206 that strategy won't
work. In any case, this way is more contained and explicit.

Fixes #150

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -442,8 +442,9 @@ class HTTPTestCase(testtools.TestCase):
             body = body.encode('UTF-8')
 
         if test['poll']:
-            count = test['poll'].get('count', 1)
-            delay = test['poll'].get('delay', 1)
+            count = int(float(self.replace_template(
+                test['poll'].get('count', 1))))
+            delay = float(self.replace_template(test['poll'].get('delay', 1)))
             failure = None
             while count:
                 try:
--- a/gabbi/tests/gabbits_intercept/poll.yaml
+++ b/gabbi/tests/gabbits_intercept/poll.yaml
@@ -26,7 +26,7 @@ tests:
 
     # Confirm that $LOCATION and $RESPONSE behave in poll.
     - name: create a thing
-      url: /poller?count=2&x=1&y=2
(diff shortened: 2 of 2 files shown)


cdent:
@josdotso and @FND if you could check over this, that would be great.

FND:
lgtm - though I'm not entirely sure what that test is really testing (likely due to lack of familiarity; I'd have to do a little more digging than I can afford right now).

More generally, I'm a little concerned about the amount of imperative (as opposed to declarative) logic in this function (or perhaps in general? that's entirely unfounded though). However, pragmatically speaking, I doubt that's actually a problem in any tangible way.

cdent:
> this function

Which one? The one that's been changed? I'm not sure at this stage of processing there's any other option than imperative. We have a value, we want to to replace_template on it, we then want to turn it into a number (or fail). This does exactly that.

(replace_template itself is pretty messy, but you should have seen it in very early versions...it's actually gotten better)

The tests are very hand wavey, they prove that the right result is happening, but don't necessarily provide clear insight into the process whereby that result is caused. It's certainly not a unit test by any stretch. But I couldn't be bothered.

FND:
D'accord.

PR #208, opened 2017-03-14
Add verbose option to `build_tests`

justanotherdot:
This allows calls to `build_tests` to accept a verbose option, making each generated verbose by default.


justanotherdot:
For the time being, I'm including checks for True and False as well as the string values since the semantics line up with that of the docs (True being synonymous to 'all'). If you'd rather it be more like the runner which only seems to accept the string values, let me know and I'll take the booleans out.

cdent:
> I'm including checks for True and False as well as the string values

That's good. I think this is good to go.

PR #210, opened 2017-03-20
Enforce type of response/content-handler test data

### LINKED ISSUE #209: response_string can lead to false positives if value is a string
The correct way to do the `response_strings` test is to provide a list of strings:

```yaml
response_strings:
    - foo
    - bar
```

If you forget and do a string, though, it can sometimes pass because the value is use as an iterable and strings iterate. In the following example of the response contains the letters `f` and `o` the test will pass:

```yaml
response_strings:
    foo
```

Stronger type checking is required.

cdent:
Each of the response and content handlers has an expected type that
the test data should be in (in the YAML file). For example
response_strings should be a list. These changes add some
enforcement of those types, without which it is possible to (rarely)
get false positives when using a str value where a list value is
expected and the assertions using that value use 'in'.

Fixes #209

### DIFF
--- a/gabbi/handlers/base.py
+++ b/gabbi/handlers/base.py
@@ -13,6 +13,9 @@
 """Base classes for response and content handlers."""
 
 
+from gabbi.exception import GabbiFormatError
+
+
 class ResponseHandler(object):
     """Add functionality for making assertions about an HTTP response.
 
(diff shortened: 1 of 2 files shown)


cdent:
Please sanity check @FND

PR #212, opened 2017-03-29
Add optional `test` param for custom content handlers

### LINKED ISSUE #211: Support for multipart/form-data encoding
`gabbi` supports sending specific binary data as single field by using `data: <@file`, but what would be the best way to replicate `multipart/form-data` requests? Something of the form:

```yaml
tests:
  - name: multipart/form-data test
    POST: /api/form-data-endpoint
    request_headers:
      content-type: multipart/form-data
    data:
      key: val
      file: <@kittens.png
```
multipart/form-data isn't entirely possible with the custom content handler approach as it requires modifying the Content-Type to have a boundary section identifier e.g. 

```http
Content-Type: multipart/form-data;boundary=---------------------------735323031399963166993862150
```

Trying to use fixtures as a hack doesn't work either as the fixtures don't have the test context, afaik. 

**Possible solutions**

1. Having some facility to plug in arbitrary python test code would be nice, but obviously it would have to 
  play nice with the current test harness being used. 

  ```yaml
    - name: multipart/form-data test
      POST: /api/form-data-endpoint
      request_headers:
        content-type: multipart/form-data
      run: extra-tests:multipart_request
  ```

2. Passing in the context to fixtures: then one can inject test specific checks and run code (e.g. run the 
    multipart_request check if test is named `X`. 

3. Implement builtin support for `multipart/form-data` encoding in a form similar to the first example above.

I've written a small proof of concept for point one but it's overkill for this. Personally I think point three should be the preferred as it should only require some minor changes in `make_one_test` and how `request` handles urrlib3 POSTs with fields given the passed content type, I believe.

justanotherdot:
This resolve #211.

`multipart/form-data` transmission requires modifying the `Content-Type` header with a `boundary` section identifier, and as such a custom content handler would not have previously had access to the running test case it was working on. This should hopefully be useful for other custom content handlers in the future.

Please let me know if the documentation could be better worded or if the check for the return of `dumps` in `assert_in_or_print_output` seems unnecessary (lines 547-551 of commit d71a477).

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -107,8 +107,11 @@ handler might accept ``application/x-www-form-urlencoded`` and
 
 If ``accepts`` is defined two additional static methods should be defined:
 
-* ``dumps``: Turn structured Python data from the ``data`` key in a
-  test into a string or byte stream.
+* ``dumps``: Turn structured Python data from the ``data`` key in a test into a
+  string or byte stream. The optional ``test`` param allows you to access the
+  current test case which may help with manipulations for custom content
+  handlers, e.g. multipart/form-data needs to add a `boundary` to the
+  Content-Type header in order to mark the appropriate sections of the body.
 * ``loads``: Turn a string or byte stream in a response into a Python data
   structure. Gabbi will put this data on the ``response_data``
   attribute on the test, where it can be used in the evaluations
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -110,8 +110,9 @@ If ``accepts`` is defined two additional static methods should be defined:
 * ``dumps``: Turn structured Python data from the ``data`` key in a test into a
   string or byte stream. The optional ``test`` param allows you to access the
   current test case which may help with manipulations for custom content
-  handlers, e.g. multipart/form-data needs to add a `boundary` to the
-  Content-Type header in order to mark the appropriate sections of the body.
+  handlers, e.g. ``multipart/form-data`` needs to add a ``boundary`` to the
+  ``Content-Type`` header in order to mark the appropriate sections of the
+  body.
 * ``loads``: Turn a string or byte stream in a response into a Python data
   structure. Gabbi will put this data on the ``response_data``
   attribute on the test, where it can be used in the evaluations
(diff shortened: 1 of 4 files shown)


gabbi/case.py
```diff
@@ -544,7 +544,11 @@ def assert_in_or_print_output(self, expected, iterable):
                 dumper_class = self.get_content_handler(self.content_type)
                 if dumper_class:
                     full_response = dumper_class.dumps(self.response_data,
-                                                       pretty=True)
+                                                       pretty=True, test=self)
+                    # Sometimes, a given content handler may not hand back a
+                    # string for transmission, so we should be careful.
+                    if not isinstance(full_response, six.string_types):
+                        full_response = str(full_response)
```
cdent: Hmmm, yeah, I'm not sure about this.

Since we're working on `response_data` here, not the `data`, and using the response content-type, not the request, the semantics of `dumps()` is a bit under-defined.

Since this change isn't specifically related to the addition of the test parameter, I'd leave this out of this pull request and perhaps register a new issue for the problem that happens here, along with a failure scenario.

It is probably something worth addressing, but in this change is a bit opportunistic rather than germane.

docs/source/handlers.rst
```diff
@@ -107,8 +107,11 @@ handler might accept ``application/x-www-form-urlencoded`` and
 
 If ``accepts`` is defined two additional static methods should be defined:
 
-* ``dumps``: Turn structured Python data from the ``data`` key in a
-  test into a string or byte stream.
+* ``dumps``: Turn structured Python data from the ``data`` key in a test into a
+  string or byte stream. The optional ``test`` param allows you to access the
+  current test case which may help with manipulations for custom content
+  handlers, e.g. multipart/form-data needs to add a `boundary` to the
```
cdent: may as well bound the content with `` to make it code

gabbi/case.py
justanotherdot: Was worried about this initially because in order to get my initial prototype to work for the multipart/form-data handler I had to pass the content along as a dict to the 'fields' param of urllib3's request. Since having context to the test means we can take advantage of properly stringifying the body and use the currently existing approach to call request (i.e. with just `body=` instead of `fields=`), I don't think even the other PR is necessary, at least not now. I'll take this out anyways.

docs/source/handlers.rst
justanotherdot: Do you mean ``multipart/form-data``, ``boundary``, or both?

cdent: Sorry, I meant to say 'content-type' and then it might have been more clear that I meant ``multipart/form-data`` (because it's a literal).

PR #214, opened 2017-04-23
Manage numerals more effectively in replacers

cdent:
Change the handling in response_replacer and jsonhandler.replace so
that numerals are preserved. The primary change here is that replace
does not always return a str: it returns the found data. The
remaining changes cascade from this fix.

* If the 'message' being replaced is fully the regex, don't do
  regex-based replacement (which requires a string). Instead return
  the new value.

* Do template replacing on test data _before_ dumping it to a string.
  This allows the small 'message' replacements described in the previous
  point.

* Assure we can recurse through lists, not just dicts, when
  replacing in data structures.

In the pull request at #206 there is a different mode of coercing
and preserving data types operates by adding a coerce method to
content handlers. While inspecting this it seemed like it might be
too complex, so this alternate option has been explored. This might
work, or perhaps a hybrid of the two.

The coerce.yaml file from #206 was copied here and adapted to fit
the different assumptions that are made in this change. Some of the
assumptions in 206 don't map well to YAML handling of literal data
formats.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -154,6 +154,9 @@ class HTTPTestCase(testtools.TestCase):
                 message[k] = self.replace_template(message[k])
             return message
 
+        if isinstance(message, list):
+            return [self.replace_template(line) for line in message]
+
         for replacer in REPLACERS:
             template = '$%s' % replacer
             method = '_%s_replace' % replacer.lower()
@@ -187,13 +190,27 @@ class HTTPTestCase(testtools.TestCase):
         return value
 
     def _environ_replace(self, message):
-        """Replace an indicator in a message with the environment value."""
+        """Replace an indicator in a message with the environment value.
+
+        If value can be a number, cast it as such. If value is a form of
+        "null", "true", or "false" cast it to None, True, False.
+        """
         value = re.sub(self._replacer_regex('ENVIRON'),
                        self._environ_replacer, message)
-        if value == "False":
+        try:
+            if '.' in value:
+                value = float(value)
+            else:
+                value = int(value)
+            return value
+        except ValueError:
+            pass
+        if value.lower() == "false":
             return False
-        if value == "True":
+        if value.lower() == "true":
             return True
+        if value.lower() == "null":
+            return None
         return value
 
     @staticmethod
@@ -351,11 +368,18 @@ class HTTPTestCase(testtools.TestCase):
         return r"%s\$%s" % (case, key)
 
     def _response_replace(self, message):
-        """Replace a content path with the value from a previous response."""
-        return re.sub(self._replacer_regex('RESPONSE'),
-                      self._response_replacer, message)
+        """Replace a content path with the value from a previous response.
 
-    def _response_replacer(self, match):
+        If the match would replace the entire message, then don't cast it
+        to a string.
+        """
+        regex = self._replacer_regex('RESPONSE')
+        match = re.match('^%s$' % regex, message)
+        if match:
+            return self._response_replacer(match, preserve=True)
+        return re.sub(regex, self._response_replacer, message)
+
+    def _response_replacer(self, match, preserve=False):
         """Replace a regex match with the value from a previous response."""
         response_path = match.group('arg')
         case = match.group('case')
@@ -368,8 +392,12 @@ class HTTPTestCase(testtools.TestCase):
         # If no handler can be found use the null replacer,
         # which returns "foo" when "$RESPONSE['foo']".
         replacer_class = replacer_class or base.ContentHandler
-        return replacer_class.replacer(
+        result = replacer_class.replacer(
             referred_case.response_data, response_path)
+        if preserve:
+            return result
+        else:
+            return six.text_type(result)
 
     def _run_request(self, url, method, headers, body, redirect=False):
         """Run the http request and decode output.
@@ -485,6 +513,7 @@ class HTTPTestCase(testtools.TestCase):
         else:
             dumper_class = self.get_content_handler(content_type)
             if dumper_class:
+                data = self.replace_template(data)
                 data = dumper_class.dumps(data, test=self)
             else:
                 raise ValueError(
--- a/gabbi/handlers/jsonhandler.py
+++ b/gabbi/handlers/jsonhandler.py
@@ -41,7 +41,7 @@ class JSONHandler(base.ContentHandler):
 
     @classmethod
     def replacer(cls, response_data, match):
-        return str(cls.extract_json_path_value(response_data, match))
+        return cls.extract_json_path_value(response_data, match)
 
     @staticmethod
     def dumps(data, pretty=False, test=None):
--- a/gabbi/tests/gabbits_intercept/coerce.yaml
+++ b/gabbi/tests/gabbits_intercept/coerce.yaml
@@ -190,3 +190,12 @@ tests:
           - $HISTORY['check posted data'].$RESPONSE['$.h.key.less_key[1]']
           - $HISTORY['check posted data'].$RESPONSE['$.h.key.less_key[2]']
         more_key: $HISTORY['check posted data'].$RESPONSE['$.h.key.more_key']
+
+- name: string internal replace
+  POST: /
+  data:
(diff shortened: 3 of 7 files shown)


cdent:
This is an alternative to #206. I'm not sure if it is better, but it feels simpler to me (which is only one dimension of "good" and only so if it actually works properly).  While looking through 206 there was so much going on that I couldn't get confident about it, so I did some experimenting of my own.

@justanotherdot if you could compare this with your version and make some comments that would be great.

@FND your eyes always appreciated.

justanotherdot:
If this is passing `coerce.yaml`, my hunch is that it'll suit the use case we use gabbi for. I'm also always for simpler solutions and it's more than likely that my pursuit of correcting edge cases landed me into a trap of over-engineering (sorry about that). That said, I want to actually test this out sometimes this week just to prove to myself if it is indeed this simple a change.

Thanks for following up with this!

cdent:
> If this is passing coerce.yaml, my hunch is that it'll suit the use case we use gabbi for.

I made a small number of adjustments between your version of coerce.yaml and mine. Most of these were related to how strings were being adapted back and forth between lists and dicts. Here's a lightly annotated diff between the two:

```diff
--- /dev/fd/63	2017-04-25 11:36:50.000000000 +0100
+++ /dev/fd/62	2017-04-25 11:36:50.000000000 +0100
@@ -23,15 +23,15 @@
   desc: data will be coerced because templates in use
   POST: /
   data:
-      one_string: $RESPONSE['$.one_string']
+      one_string: !!str "$RESPONSE['$.one_string']"
# this is just an experiment with yaml type forcing
       one_int: $RESPONSE['$.one_int']
       one_float: $RESPONSE['$.one_float']
   response_json_paths:
-      $.one_string: 1
+      $.one_string: "1"
# this is actually what we want because we sent a string we want to
# verify it
       $.one_int: 1
       $.one_float: 1.1
   response_strings:
-      - '"one_string": 1'
+      - '"one_string": "1"'
# same
       - '"one_int": 1'
       - '"one_float": 1.1'
 
@@ -53,9 +53,9 @@
   response_json_paths:
       $[0]: 1
       $[1]: 2
-      $[2]: 3
+      $[2]: "3"
# same
   response_strings:
-      - '[1, 2, 3]'
+      - '[1, 2, "3"]'
# same
 
 - name: object with list
   desc: without recursive handling no coersion
@@ -72,16 +72,15 @@
       $.collection[1].alpha: 2
       $.collection[1].beta: climb
   response_strings:
-      - "alpha": 1
-      - "beta": "max"
-      - [1, 2]
+      - '"alpha": 1'
+      - '"beta": "max"'
# i'm not sure how/where [1,2] gets into it, so I just took it out
 
 - name: post extra data
   POST: /
   data:
     a: 1
     b: 1.0
-    c: [1,2,3]
+    c: '[1,2,3]'
# this is sending a string that looks like an array, and should stay
# one
     d: true
     e: false
     f:
@@ -91,11 +90,10 @@
       key:
         less_key: [1, true, null]
         more_key: 1
-    i: ∧
# the unicode handling in env variables is weird enough that I took
# out that stuff
   response_json_paths:
     a: 1
     b: 1.0
-    c: [1,2,3]
+    c: '[1,2,3]'
# see above
     d: true
     e: false
     f:
@@ -105,7 +103,6 @@
       key:
         less_key: [1, true, null]
         more_key: 1
-    i: ∧
 
 - name: Check posted data
   POST: /
@@ -118,11 +115,10 @@
     f: $RESPONSE['$.f']
     g: $RESPONSE['$.g']
     h: $RESPONSE['$.h']
-    i: $RESPONSE['$.i']
   response_json_paths:
     a: 1
     b: 1.0
-    c: [1,2,3]
+    c: '[1,2,3]'
     d: true
     e: false
     f:
@@ -132,7 +128,6 @@
       key:
         less_key: [1, true, null]
         more_key: 1
-    i: ∧
 
 - name: Post again and check the results
   POST: /
@@ -145,7 +140,6 @@
     f: $HISTORY['post extra data'].$RESPONSE['$.f']
     g: $HISTORY['post extra data'].$RESPONSE['$.g']
     h: $HISTORY['post extra data'].$RESPONSE['$.h']
-    i: $HISTORY['post extra data'].$RESPONSE['$.i']
   response_json_paths:
     a: $ENVIRON['ONE']
     b: $ENVIRON['DECIMAL']
@@ -162,7 +156,6 @@
           - $ENVIRON['TRUE']
           - $ENVIRON['NULL']
         more_key: $ENVIRON['ONE']
-    i: $ENVIRON['UNICODE']
 
 - name: Post again and check the results (reversed)
   POST: /
@@ -182,7 +175,6 @@
           - $ENVIRON['TRUE']
           - $ENVIRON['NULL']
         more_key: $ENVIRON['ONE']
-    i: $ENVIRON['UNICODE']
   response_json_paths:
     a: $HISTORY['Check posted data'].$RESPONSE['$.a']
     b: $HISTORY['Check posted data'].$RESPONSE['$.b']
@@ -198,6 +190,3 @@
           - $HISTORY['Check posted data'].$RESPONSE['$.h.key.less_key[1]']
           - $HISTORY['Check posted data'].$RESPONSE['$.h.key.less_key[2]']
         more_key: $HISTORY['Check posted data'].$RESPONSE['$.h.key.more_key']
-    i: $HISTORY['Check posted data'].$RESPONSE['$.i']
```

justanotherdot:
IIRC (I may be a bit off here but the rough idea is still the same, I believe) one odd scenario I think this PR resolves is when you have a string that needs to have a replacement made inside of it, e.g.

```yaml
POST: /api/0.1/item/delete/$RESPONSE['$.id']
```

Which would then generate a string looking like...

```python
"/api/0.1/item/delete/$RESPONSE['$.id']"
```
Of which the value would be looked up and replaced, but be enclosed in quotes based on the fact that it was being `str`ed.

```python
/api/0.1/item/delete/"1"
```

I have added a test case that checks for this, though, with this diff patched in and it seems to pass; here's the test (added at the end of `coerce.yaml`):

```yaml
- name: String internal replace
  POST: /
  data:
    endpoint_resp: /api/0.1/item/$HISTORY['check posted data'].$RESPONSE['$.a']
    endpoint_var: /api/0.1/item/$ENVIRON['ONE']
  response_json_paths:
    endpoint_resp: /api/0.1/item/1
    endpoint_var: /api/0.1/item/1
```

This should theoretically be the only thing that would have caused problems in the tests I'm running, but I'll still need to find the time this week to run them and verify that there's no other edge cases I've overlooked.

justanotherdot:
Just ran those tests I mentioned and this seems to work just fine :+1:

cdent:
> this seems to work just fine

Yay!

I'll add your test, looks useful.

Now just need to get @FND to confirm or deny and we'll make it go.

gabbi/case.py
```diff
@@ -187,13 +190,27 @@ def _clean_query_value(self, value):
...
+        """Replace an indicator in a message with the environment value.
+
+        If value can be a number, cast it as such. If value is a form of
+        "null", "true", or "false" cast it to None, True, False.
+        """
         value = re.sub(self._replacer_regex('ENVIRON'),
                        self._environ_replacer, message)
-        if value == "False":
+        try:
+            if '.' in value:
```
FND: This doesn't seem very Pythonic; why not
```python
try:
    value = int(value)
except ValueError:
    try:
        value = float(value)
    except ValueError:
        pass
```

gabbi/case.py
```diff
@@ -187,13 +190,27 @@ def _clean_query_value(self, value):
...
+            return value
+        except ValueError:
+            pass
+        if value.lower() == "false":
             return False
-        if value == "True":
+        if value.lower() == "true":
             return True
+        if value.lower() == "null":
+            return None
```
FND: This seems like we're reinventing JSON/YAML parsing; do we really wanna do that?

PS: Would it be silly to ask for caching `value.lower()` in a local variable?

cdent: this is the value of the environment variable being parsed and all environment variables are strings but in this case we want them have stronger types than that so they can actually be used as those types in things like json.

and yes, caching `value.lower()` is silly. not completely silly, but silly

PR #215, opened 2017-07-06
Template expansion in LHS of json path test

### LINKED ISSUE #170: substitutions not allowed in left hand side of json paths in response_json_paths
``` yaml
$.allocations.$ENVIRON['RP_UUID'].DISK_GB: 40
```

does not work, maybe it should?

cdent:
This change implements template substitution ($ENVIRON, $RESPONSE,
etc) in the left hand side of a response_json_paths test. This enables
testing in situations where responses may include a server side
generated identifier as the key in an object structure.

As things stand, this is the bare minimum implementation that assumes
that the test creator will be responsible for dealing with the
vargrancies of quoting that will be required in order to generate
legitimate JSON path expressions. For example if there is a uuid (with
hyphens) at $RESPONSE['$.id'] then this expression is likely to fail:

    $.nested.structure.$RESPONSE['$.id'].name: foobar

as it will evaluate to something like

    $.nested.structure.ADC8AAFC-D564-40D1-9724-7680D3C010C2.name: foobar

which _may_ be treated as an arithemtic expression by the json parser.
The test creator would need to do:

    $.nested.structure["$RESPONSE['$.id']"].name: foobar

or variants thereof to get a useful expression.

Docs to follow if this is considered an okay way to implement the
feature.

Fixes #170

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -274,7 +274,8 @@ All of these variables may be used in all of the following fields:
 * ``data``
 * ``request_headers``
 * ``response_strings``
-* ``response_json_paths`` (on the value side of the key value pair)
+* ``response_json_paths`` (in both the key and value, see
+  :ref:`json path substitution <json-subs>` for more info)
 * ``response_headers`` (on the value side of the key value pair)
 * ``response_forbidden_headers``
 * ``count`` and ``delay`` fields of ``poll``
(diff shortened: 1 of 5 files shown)


cdent:
@jd and @jaypipes you've thought about this topic before. If you've got input that would be great.

PR #216, opened 2017-07-07
Read full json

cdent:
The change allows some or all of a JSON value to be read from disk to be used in a json path test. Examples are are in the `data.yaml` test file and in the `jsonpath.rst` file. This will help with test that either need to validate full responses or that have complex responses that need to be verified but are big enough that having them in the yaml file muddies the water.

This feature has been requested by a few different parties. It's not something I originally wanted to do but enough people wanted it and it is easy enough to do that...why fight it?

/cc @FND @jasonamyers @tomviner @justanotherdot

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -174,6 +174,10 @@ class HTTPTestCase(testtools.TestCase):
 
         return message
 
+    def load_data_file(self, filename):
+        """Read a file from the current test directory."""
+        return self._load_data_file(filename)
+
     def _assert_response(self):
         """Compare the response with expected data."""
         self._test_status(self.test_data['status'], self.response['status'])
@@ -505,7 +509,7 @@ class HTTPTestCase(testtools.TestCase):
         """
         if isinstance(data, str):
             if data.startswith('<@'):
-                info = self._load_data_file(data.replace('<@', '', 1))
+                info = self.load_data_file(data.replace('<@', '', 1))
                 if utils.not_binary(content_type):
                     data = six.text_type(info, 'UTF-8')
                 else:
--- a/gabbi/handlers/jsonhandler.py
+++ b/gabbi/handlers/jsonhandler.py
@@ -14,6 +14,8 @@
 
 import json
 
+import six
+
 from gabbi.handlers import base
 from gabbi import json_parser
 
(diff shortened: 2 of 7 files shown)


jaypipes:
Good thing here. I like it. A lot. Will make some of the messier JSONPath entries in Placement gabbits a thing of the past. ++ from me! :)

gabbi/handlers/jsonhandler.py
```diff
@@ -86,6 +88,13 @@ def action(self, test, path, value=None):
         except ValueError:
             raise AssertionError('json path %s cannot match %s' %
                                  (path, test.response_data))
+
+        # read data from disk if the value starts with '<@'
+        if isinstance(value, str) and value.startswith('<@'):
+            info = test.load_data_file(value.replace('<@', '', 1))
```
jd: value[2:] no? whatever :p

gabbi/tests/gabbits_intercept/data.yaml
```diff
@@ -49,3 +49,12 @@ tests:
       request_headers:
           content-type: text/plain
       data: <@utf8.txt
+
+    - name: json value from disk
+      POST: /
+      request_headers:
+          content-type: application/json
+      data: <@data.json
```
jd: why is not this file in the commit?

gabbi/handlers/jsonhandler.py
cdent: I don't like using slices like that, it makes me have to decode the slice grammar whereas what I've got above is visible and readable.

FND: FWIW, I think @jd's suggestion would be more readable ¯\\\_(ツ)_/¯

gabbi/case.py
```diff
@@ -174,6 +174,10 @@ def replace_template(self, message):
 
         return message
 
+    def load_data_file(self, filename):
+        """Read a file from the current test directory."""
+        return self._load_data_file(filename)
```
FND: This seems like a bit of weird method; what's the point?

gabbi/handlers/jsonhandler.py
EdLeafe: Also prefer @jd's approach, especially since the line above uses 'startswith', making the intent clear.

PR #221, opened 2017-10-09
use_prior_test directive and flag

dkt26111:
* Added use_prior_test flag to build_tests.  Defaults to True.
* Added use_prior_test directive in YAML file.
  - If not set in the YAML file, it will default to the value set by
    build_tests.
  - If it is set in the YAML file, its value will override the
    use_prior_test value set by build_tests.
* When executing a test, only run prior test if use_prior_test is True.
* Added tests verifying build_tests and HTTPTestCase test_request
  behavior with use_prior_test.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -135,8 +135,8 @@ class HTTPTestCase(testtools.TestCase):
         if self.test_data['skip']:
             self.skipTest(self.test_data['skip'])
 
-        if self.prior and not self.prior.has_run and \
-                self.test_data['use_prior_test']:
+        if (self.prior and not self.prior.has_run and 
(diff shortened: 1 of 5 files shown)


dkt26111:
I'm abandoning this pull request to fix the pep8 issue.

cdent:
First glance this looks great, but I'll need to look at it properly when I have some brains.  I suspect we'll have some rounds on the name too, not got any ideas yet.

cdent:
I'll get the docs in a followup

dkt26111:
Thanks everyone for the reviews.  Also thanks for taking care of the docs @cdent .

PR #222, opened 2017-10-10
Various regular expression fixes

zaneb:
This series fixes a bunch of issues with matching against regular expressions when:

* The output data contains non-ASCII characters
* The regex itself begins or ends with a `/`
* The expression is just a literal single `/`, not a regex
* The expression is a substitution and the substituted values begins and ends with a `/`
* A substitution occurs inside the regex and the substituted data contains characters that have special meaning in a regular expression

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -147,15 +147,17 @@ class HTTPTestCase(testtools.TestCase):
                 return handler
         return None
 
-    def replace_template(self, message):
+    def replace_template(self, message, escape_regex=False):
         """Replace magic strings in message."""
         if isinstance(message, dict):
             for k in message:
-                message[k] = self.replace_template(message[k])
+                message[k] = self.replace_template(message[k],
+                                                   escape_regex=escape_regex)
             return message
 
         if isinstance(message, list):
-            return [self.replace_template(line) for line in message]
+            return [self.replace_template(line, escape_regex=escape_regex)
+                    for line in message]
 
         for replacer in REPLACERS:
             template = '$%s' % replacer
@@ -163,7 +165,8 @@ class HTTPTestCase(testtools.TestCase):
             try:
                 if template in message:
                     try:
-                        message = getattr(self, method)(message)
+                        replace = getattr(self, method)
+                        message = replace(message, escape_regex=escape_regex)
                     except (KeyError, AttributeError, ValueError) as exc:
                         raise AssertionError(
                             'unable to replace %s in %s, data unavailable: %s'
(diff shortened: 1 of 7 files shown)


cdent:
Thanks, this all looks very good, but I'm too tired to merge it tonight. @FND or @jasonamyers may have some comments; otherwise, I'll give it a proper look in the morning.

As you could probably tell the regex support was very rudimentary, thrown in as a "good enough". Was there a particular thing that started you needing to the fix this? The unicode handling?

gabbi/handlers/jsonhandler.py
```diff
@@ -94,9 +94,11 @@ def action(self, test, path, value=None):
 
         expected = test.replace_template(value)
         # If expected is a string, check to see if it is a regex.
-        if (hasattr(expected, 'startswith') and expected.startswith('/')
-                and expected.endswith('/')):
-            expected = expected.strip('/').rstrip('/')
+        if (isinstance(expected, six.string_types)
```
cdent: The reason I used `hasattr` is because I have some kind of bias against `isinstance` that I should probably just get over.

gabbi/handlers/jsonhandler.py
```diff
@@ -94,9 +94,11 @@ def action(self, test, path, value=None):
...
+                and expected.startswith('/')
+                and expected.endswith('/')
+                and len(expected) > 1):
```
cdent: Ooops. Good catch.

gabbi/case.py
```diff
@@ -193,14 +196,28 @@ def _clean_query_value(self, value):
         value = value.encode('UTF-8')
         return value
 
-    def _environ_replace(self, message):
+    @staticmethod
+    def _regex_replacer(replacer, escape_regex):
```
cdent: Hmmm, yeah, I get why this is required, but the added indirection is a bummer.

zaneb:
I'm not actually blocked on any of this. I was spelunking in the code looking for a way to fix https://bugs.launchpad.net/heat/+bug/1721692. Saw some issues. Fixed 'em.

gabbi/case.py
zaneb: It'd be possible to rewrite all of the _replacer() functions to take an `escape_regex` argument and use `functools.partial()` instead, but I figured that wasn't actually any better.

gabbi/handlers/jsonhandler.py
zaneb: Maybe reading https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__ will help you get over it ;)

PR #225, opened 2017-10-25
Allow to load python object from yaml

sileht:
It can be interesting to write custom object to compare values.

For example, I need to ensure an output is equal to .NAN

Because .NAN == .NAN always returns false. We currently can't compare it
with assert_equals().

With the unsafe yaml loader we can register a custom method to check
NAN, for example:

    class IsNAN(object):
        @classmethod
        def constructor(cls, loader, node):
            return cls()

        def __eq__(self, other):
            return numpy.isnan(other)

    yaml.add_constructor(u'!ISNAN', ISNAN.constructor)

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -89,7 +89,8 @@ def run():
     if not input_files:
         success = run_suite(sys.stdin, handler_objects, host, port,
                             prefix, force_ssl, failfast,
-                            verbosity=verbosity)
+                            verbosity=verbosity,
+                            safe_yaml=args.safe_yaml)
         failure = not success
     else:
         for input_file in input_files:
@@ -99,7 +100,8 @@ def run():
                 success = run_suite(fh, handler_objects, host, port,
                                     prefix, force_ssl, failfast,
                                     data_dir=data_dir,
-                                    verbosity=verbosity, name=name)
+                                    verbosity=verbosity, name=name,
+                                    safe_yaml=args.safe_yaml)
             if not success:
                 failures.append(input_file)
             if not failure:  # once failed, this is considered immutable
@@ -114,9 +116,10 @@ def run():
 
 
 def run_suite(handle, handler_objects, host, port, prefix, force_ssl=False,
-              failfast=False, data_dir='.', verbosity=False, name='input'):
+              failfast=False, data_dir='.', verbosity=False, name='input',
+              safe_yaml=True):
     """Run the tests from the YAML in handle."""
-    data = utils.load_yaml(handle)
+    data = utils.load_yaml(handle, safe=safe_yaml)
     if force_ssl:
         if 'defaults' in data:
             data['defaults']['ssl'] = True
@@ -223,6 +226,14 @@ def _make_argparser():
         dest='verbosity',
         choices=['all', 'body', 'headers'],
         help='Turn on test verbosity for all tests run in this session.'
+    )
+    parser.add_argument(
+        '--unsafe-yaml',
+        dest='safe_yaml',
+        action='store_false',
+        default=True,
+        help='Turn on recognition of Python objects in addition to '
+             'standard YAML tags.'
     )
     return parser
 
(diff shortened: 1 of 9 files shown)


cdent:
Seems reasonable but I'm having a hard time visualizing. Have you got an example where you're already using it?

I suppose since we're in a controlled environment there's no reason to insisnt on safe_load.

sileht:
I plan to use this for: https://github.com/gnocchixyz/gnocchi/pull/446/files#diff-f12b9f5f43a3ce0d98cc7e1dac0cee7fR432

sileht:
I plan to replace .NAN by !ISNAN to make the  test pass

cdent:
Ah, okay. Thanks that makes it a lot more clear. If you could add a section to the faq.rst about how to make use of this that would be great, but if you don't get to it before I get to merging it (sometime tomorrow, going to bed now) I'll add something. Thanks.

FND:
I understand the idea - it's a good one which might open up interesting new avenues - but I'm pretty uncomfortable with this assumption:

> I suppose since we're in a controlled environment there's no reason to insisnt on safe_load.

I'd much rather individual suites opt into that.

cdent:
> I'd much rather individual suites opt into that.

So your preference would be that the test runner set a flag or something of that ilk which would switch away from using `safe_load`? That might be useful, but it's probably also worthwhile to bone up on the difference between `safe_load` and `load` (I haven't yet because of time).

cdent:
For reference, some conversations with FND about this starts at: http://p.anticdent.org/1AsR

We have four options for moving forward:

1. merge this as is
0. add a flag to build_tests to switch in load for safe_load upon request
0. environ var that does same
0. reject this

4 is not a good option.

Will come back to this asap.

sileht:
I prefer 1. and 2.

FND:
I guess 2. (plus gabbi-run support) would (more or less) address my (still fairly vague and unarticulated) concerns - though I'm not actually sure about that. ¯\_(ツ)_/¯ Sorry I can't be more helpful at this time.

sileht:
I'm OK with 2, I just need a name for the new parameter. safe_yaml_loading ? unsafe_yaml_loading ? unsafe_yaml ? allow_custom_yaml_symbol ? enable_custom_symbol ?

cdent:
How about `safe_yaml`, defaults to `True`, and callers can set it to`False` if they'd like otherwise?

FND:
After (yet another) [lengthy IRC discussion](http://p.anticdent.org/2tXU) about my concerns (mostly about separation of concerns: dict-style data model vs. YAML representation), cdent has coerced my blessing (not that that was mine to give; the best I could hope for was more filibustering).

So as far as I'm concerned, `build_tests(…, safe_yaml=False)` seems fine. (I briefly tried to find a more salient term in places like [tenderlove's write-up](http://tenderlovemaking.com/2013/02/06/yaml-f7u12.html), but failed to come up with anything more alarming.)

sileht:
Pull-request updated, HEAD is now d48213c7ffe2a03d37db2e1b6698736b4b48d265

gabbi/runner.py
```diff
@@ -224,6 +227,14 @@ def _make_argparser():
         choices=['all', 'body', 'headers'],
         help='Turn on test verbosity for all tests run in this session.'
     )
+    parser.add_argument(
+        '--unsafe-yaml',
+        dest='safe_yaml',
+        action='store_false',
+        default=True,
+        help='Trun on recognization of Python object in additional of '
```
cdent: Turn on recognition of Python objects in addition to standard YAML tags

sileht:
Pull-request updated, HEAD is now e0114ea4c8d3bde5988facfcfbc5d0601365761c

sileht:
Pull-request updated, HEAD is now 3fcf2ce3d27da13af5414ffed13858d1fbdcc8dd

gabbi/runner.py
sileht: done

sileht:
Pull-request updated, HEAD is now 98adca65e05b7de4f1ab2bf90ab0521af5030f35

PR #229, opened 2017-12-08
Fix pytest handling of fixture-level skips

### LINKED ISSUE #228: when a fixture start_fixture raise unittest.SkipTest, this first test in suite is marker skip，but the other tests continue to run
when I  user pytest  to run a yaml file， I hope when start_fixture raise unittest.SkipTest， this all tests in this yaml should marked skip， but  now only this first test marked skip， and the other tests continue to run。

cdent:
Due to an oversight when developing the pytest functionality, the
feature wherein if a fixture raised a SkipTest exception, all the
tests within the suite would be skipped did not work.

It turns out this is because several bad assumption were being made:

* That the start_fixture had access to the list of tests in
  that pytest will run. It doesn't. It has access to a list
  of tests prior to the tests being process by the pytester
  plugin. These are not the tests that will be run and
  manipulating them to be skipped will not work.

  Therefore in the ptester plugin pre-processing the list of
  real tests is added to the information the start_fixture is
  aware of.

* Calling addSkip on the pytest result class does not return.
  What needs to happen instead is that a skip marker is applied
  to the tests. Since we now have all the tests as described in
  the bullet above, it is possible to traverse them and mark
  them.

This was all tested against the live tests, which will skip
if GABBI_SKIP_NETWORK is true. In the process, it became clear
that there is no need to have different test loader files for
pytest versus unit tests. The load_tests and pytest_generate_tests
methods can live in the same file and will only be used by
the test loaders that want them.

So now all the YAML files are run by both test running, increasing
the coverage provided when using pytest.

Fixes #228


cdent:
This is a bit noisy, under normal circumstances I would have said to do this in different patches, but it took quite a bit of messing around to prove to myself that this was working, and once I had done all that I didn't want to lose it nor spend any more time.

cdent:
@sileht you might be interested in this

gabbi/tests/test_inner_fixture.py
```diff
@@ -21,6 +21,8 @@
 import fixtures
 
 from gabbi import driver
+# TODO(cdent): this test_* needs to be imported bare or things do not work
```
FND: FWIW, I don't understand this note.

cdent: Yeah, that's a copied note from the first time I was working through the pytest stuff. I should probably make it more clear. Basically pytest won't use the generator if the imported symbol is not in the module. The TODO-ness of it is bascially supposed to mean "ugh, figure out something to make this less gross."

All the pytest stuff is a big messy hack, it still awaits someone who knows what they are doing to come along, point and laugh, and then fix it.

gabbi/pytester.py
```diff
@@ -90,14 +90,18 @@ def a_pytest_collection_modifyitems(items, config):
         if cleanname.startswith('start_'):
             test = item.callspec.params['test']
             result = item.callspec.params['result']
-            STARTS[suitename] = (test, result)
+            # TODO(cdent): Consider a named tuple here
+            STARTS[suitename] = (test, result, [])
```
jasonamyers: I wished it was a named tuple as well.

Issue #231, opened 2017-12-14
Substitution not occurring for header key


joshleeb:
Substitution doesn't seem to occur for the key of `request_headers`. For example:

Having this test works as expected, where `$ENVIRON['HEADER_VAL']` is replaced by the value of `HEADER_VAL`.

```yaml
- name: Working test
  desc: This is a working test
  GET: /api/some/endpoint
  request_headers:
    Header-Key: $ENVIRON['HEADER_VAL']
  status: 200
```

Yet, this test doesn't work as `$ENVIRON['HEADER_KEY']` isn't substituted with `HEADER_KEY`.

```yaml
- name: Failing test
  desc: This is a failing test
  GET: /api/some/endpoint
  request_headers:
    $ENVIRON['HEADER_KEY']: $ENVIRON['HEADER_VAL']
  status: 200
```

As a side note, this may have some relevance to #219.

cdent:
Thanks for noticing this and for providing #232. It's mostly something that wasn't done initially because there was no immediate use case. I suspect I was also maintaining some kind of bias against unpredictable headers.

Out of curiosity, what's your use case? What are the headers that you have that aren't predictable?

joshleeb:
I use a wrapper around gabbi that, amongst other things, helps to handle request authentication throughout the suite of e2e tests. Specifically, this wrapper sets variables in `$ENVIRON` which are read from a config file. It's then possible to dynamically provide the required authentication token in a request header.

Up until recently, the header key hasn't changed, but now I am needing to extend the wrapper to read the header key from the config as well so it can by dynamically changed in the tests. 

Overall it provides a bit more flexibility with the test suite in being able to run it with different authentication methods and tokens.

PR #232, opened 2017-12-14
Run replace template for header keys

joshleeb:
Ref. Substitution not occurring for header key #231 

Run `replace_template` for header keys as well as header values.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -489,12 +489,18 @@ class HTTPTestCase(testtools.TestCase):
 
         method = test['method'].upper()
         headers = test['request_headers']
+
+        replaced_headers = {}
         for name in headers:
             try:
-                headers[name] = self.replace_template(headers[name])
+                replaced_name = self.replace_template(name)
+                replaced_headers[replaced_name] = self.replace_template(
+                    headers[name]
+                )
             except TypeError as exc:
                 raise exception.GabbiFormatError(
                     'malformed headers in test %s: %s' % (test['name'], exc))
+        headers = replaced_headers
 
         if test['data'] != '':
             body = self._test_data_to_string(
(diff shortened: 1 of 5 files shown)


joshleeb:
I've added a test and checked that it will fail without the patch applied in this PR. 

Do you have any suggestions of what to add to the documentation? The current docs already seem to describe the added variable substitution:

> The variables are replaced with real values during test processing.

cdent:
Sorry, for not being explicit, I was thinking in format.rst near the "All of these variables may be used in all of the following fields:" section, adding a short to the request headers description.

But also, when looking at that, I realized you've implemented this for just request_headers, which is sufficient, and worth merging, but if you're feeling up to it, using replace template on the response_headers header name might be useful as well. If you've already spent your personal time budget as much as you want on this, don't worry about it, I'll get to it asap. Having it balanced like that would be good.

Thanks again. I'm glad gabbi's useful for you.

gabbi/case.py
```diff
@@ -487,21 +502,16 @@ def _run_test(self):
         self.url = base_url
         full_url = self._parse_url(base_url)
 
+        # Replace variables in headers with variable values. This includes both
+        # in the header key and the header value.
+        test['request_headers'] = self._replace_headers_template(
+            test['name'], test['request_headers'])
+        test['response_headers'] = self._replace_headers_template(
```
cdent: I'll need to look at this more closely in the daylight, but there's a chance the response half of this should be in `HeadersResponseHandler`, but maybe not. I've got insufficient brain at the moment to decide.

cdent:
I'm still not completely decide about whether this shouldn't eventually be processed in the handler, but this is okay as is, adds a good feature, and doesn't preclude other ways of doing it in the future, so: yeah.

PR #233, opened 2017-12-15
Add tox to test requirements

joshleeb:
I came across this during the development of #232. However, I thought it wasn't relevant so I would add it to a separate PR.


cdent:
A while back, someone made a pull request to remove tox from test-requirements.txt in  30649f6c977ab68b8fb546ea5d350949acf6227e

The rationale at the time, I think, was that unless you already had tox you wouldn't engage test-requirements.txt anyway.

I guess it depends on how things are ordered and what's doing the automation. What was your path to discovering the problem?

joshleeb:
I first created a new virtualenv and installed `gabbi` with:

```bash
$ pip install -e . -r requirements.txt -r test-requirements.txt
```

Then, to run the tests, I noticed there was a `tox.ini` file that was setup to run tests, so I tried to run `tox` as a quick way to run the tests. And that's where I hit this snag. 

I also found it useful to run `tox` before submitting a PR to know, with reasonable certainty, if the CI jobs will pass.

It isn't really a big deal but I imagine for most people who setup a new virtualenv to contribute to `gabbi` they would run into this issue.

cdent:
> It isn't really a big deal but I imagine for most people who setup a new virtualenv to contribute to gabbi they would run into this issue.

I think you're probably right that having it is useful. I'm cc'ing @FND and @jasonamyers to see if they have any memory of when it went away.

jasonamyers:
It went away exactly like you mentioned. Often rod is installed at the user of system level no need to reinstall it during testing as you be reinstalling tox into each tox test environment tox. It would make sense to have tox in requirements or requirements-dev.txt

jasonamyers:
Often “tox” is installed not rod

FND:
Unfortunately, I don't have _any_ memory. Nevertheless, as what amounts to a casual contributor at this point, I would prefer gabbi's dependency declarations to be comprehensive, without relying on global/system-level packages - i.e. I would expect pretty much what @joshleeb described. (I vaguely remember @cdent having to guide me through the setup process more than once - which neither of us particularly enjoyed.)

cdent:
It's always the simplest things that are complex. 🤦‍♂️  It appears that requirements-dev.txt is relatively common, should we add it and put tox in there instead of test-requirements.txt?

joshleeb:
That seems reasonable to me. Maybe name it `dev-requirements.txt` to keep consistency with `test-requirements.txt`?

That way it can be an option not to install `tox` (or not to install it again) and if a contributor decides to run `pip` with all the `*requirements.txt` files then they would be fine there too.

jasonamyers:
It’s definitely more common to have requirements.txt and requirements-dev.txt

jasonamyers:
You shouldn’t be installing test-requirements that should be used by tox only. Since that’s the desired testing routing

joshleeb:
I'd be happy to update the PR to split requirements into `requirements.txt` and `requirements-dev.txt`?

I'm thinking `requirements.txt` wouldn't change. Then `test-requirements.txt` would become `requirements-dev.txt` and include `tox`.

cdent:
> I'm thinking requirements.txt wouldn't change. Then test-requirements.txt would become requirements-dev.txt and include tox.

Sorry, slightly more complex than that:

* requirements.txt stays as is (the things required for gabbi to run)
* test-requirements.txt for things required in environments created by tox
* requirements-dev.txt for things required to set up a dev env where it is possible to then run the tests. This _might_ be just tox?

I've got a bad cold, I'm not processing information well.

jasonamyers:
So Tox needs a set of requirements, and in my past experience that does not include tox. So we'd have requirements.txt for running it, requirements-dev.txt to have all the things developers need, and test-requirements.txt for all the tox needs. I'd imagine requirements-dev.txt to -r test-requirements.txt and include tox.

requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+-r test-requirements.txt
```
cdent: @jasonamyers can you explain (in more detail) why this should be in here? I thought we were just trying to get to an environment where it is possible to run the tests (with tox)?

Sorry for belaboring this, it is starting to go from hilarious to depressing, but if we're going to change it, I'd like to understand it and get it right.

jasonamyers: It depends on how you want the devs to work. Is tox the source of testing truth? if so, then we can dismiss this line. If not, then this will give them a virtualenv with all the tools in it.

cdent: > Is tox the source of testing truth?

In gabbi itself, yes.

jasonamyers: for a developer when their setup their development environment what do you want them to have in it?

cdent: That is the question I was trying to get an answer for earlier; whether there is anything beyond tox that's required. So I guess I'll just try it. One moment... tox, git, and an editor.

jasonamyers:
:+1: then good to remove it

cdent:
Thanks for sticking with this @joshleeb . I'll make some tweaks in the docs to mention its existence before I package up the release that includes the header changes you made (thanks very much for those too).

PR #240, opened 2018-01-06
Pass strict=True when creating httpclient

### LINKED ISSUE #239: Test for status code 200 succeeds for invalid HTTP responses
Using a simple test like:

    tests:
       - name: retrieve root
         GET: /
         status: 200

With server:

`echo "HTTP/1.0 404 Not Found" | nc -l 8000`

You'll get the expected error:

`'404' not in ['200']`

However, an invalid HTTP response like this will succeed:

`echo "X/1.0 404 Not Found" | nc -l 8000`

It seems like any response (even empty) other than an actual valid error status line will succeed when testing for status 200.

When testing for other status codes invalid responses aren't accepted, so it seems 200 is a special case.

cdent:
When strict=True, urllib3's PoolManager (and the underlying
ConnectionPool and HTTPConnection) require that the first line
received in an HTTP response be a valid HTTP status line. If
not a BadStatusLine exception is raised. The actual handling
of strictness is within the httplib and http.lib modules
(depending on Python version). In Python 3, strict is ignored
because the library is always strict.

If strict=False (the default) a bad status line will be ignored
and, bizarrely, the status of the response is set to 200,
misleading gabbi. This is because a bad status line is
interpreted as possibly being an HTTP 0.9 response.

Since gabbi has no interest in HTTP 0.9, nor in being mislead
by inaccurate status values, always call strict when creating
the client.

Fixes #239

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -182,5 +182,5 @@ def get_http(verbose=False, caption=''):
         if verbose == 'headers':
             body = False
         return VerboseHttp(body=body, headers=headers, colorize=colorize,
-                           stream=stream, caption=caption)
-    return Http()
+                           stream=stream, caption=caption, strict=True)
+    return Http(strict=True)


cdent:
I didn't do any test for this since the issue is outside of gabbi. All any test would do in this case (unless it had mocks of external libraries) is confirm that python is being called as written, and I think we can trust Python in this case.

cdent:
I’ve decided there’s a cleaner way, but it’ll have to wait until I’m back at my puter. Same concept but less duplication.

gabbi/httpclient.py
```diff
@@ -182,5 +182,5 @@ def get_http(verbose=False, caption=''):
         if verbose == 'headers':
             body = False
         return VerboseHttp(body=body, headers=headers, colorize=colorize,
-                           stream=stream, caption=caption)
-    return Http()
+                           stream=stream, caption=caption, strict=True)
+    return Http(strict=True)
```
FND: FWIW, while I'm usually a proponent of return-early abortion, in this case it actually confused me because
a) the abort condition, and thus relevant context, didn't fit into the diff (or mental buffer, for that matter), so I was briefly puzzled as to why there were two return statements at different levels of indentation
b) the abort block is significantly _more_ complex than the regular path

Thus I'd either flip the condition (with some reluctance because negotiation is a little awkward) despite some hesitation or use `else` (though (b) would still apply there).

cdent:
Changed my mind, will do the cleaner way later.

gabbi/httpclient.py
cdent: Feel free to make that adjustment in a later patch. I don't want to do so in this change.

FND: Makes sense - which do we prefer:
```python
if not verbose:
    return return Http(strict=True)

headers = False if verbose == 'body' else True
body = False if verbose == 'headers' else True
return VerboseHttp(…)
```
or
```python
if verbose:
    headers = False if verbose == 'body' else True
    body = False if verbose == 'headers' else True
    return VerboseHttp(…)
else:
    return return Http(strict=True)
```

cdent: No strong preference, but probably the former a bit more, because of less indentation. Thanks.

PR #241, opened 2018-01-07
Simplify HTTP setup for readability

FND:
cf. https://github.com/cdent/gabbi/pull/240#discussion_r160033154

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -172,15 +172,10 @@ class VerboseHttp(Http):
 
 def get_http(verbose=False, caption=''):
     """Return an Http class for making requests."""
-    if verbose:
-        body = True
-        headers = True
-        colorize = True
-        stream = sys.stdout
-        if verbose == 'body':
-            headers = False
-        if verbose == 'headers':
-            body = False
-        return VerboseHttp(body=body, headers=headers, colorize=colorize,
-                           stream=stream, caption=caption, strict=True)
-    return Http(strict=True)
+    if not verbose:
+        return Http(strict=True)
+
+    headers = False if verbose == 'body' else True
+    body = False if verbose == 'headers' else True
+    return VerboseHttp(headers=headers, body=body, stream=sys.stdout,
+                       caption=caption, colorize=True, strict=True)
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -10,7 +10,6 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""Subclass of Http class for verbosity."""
 
 from __future__ import print_function
 
@@ -29,10 +28,10 @@ urllib3.disable_warnings()
 
 
 class Http(urllib3.PoolManager):
-    """A subclass of the urllib3.PoolManager to munge the data.
+    """A subclass of the ``urllib3.PoolManager`` to munge the data.
 
     This transforms the response to look more like what httplib2
-    provided when it was used as the httpclient.
+    provided when it was used as the HTTP client.
     """
 
     def request(self, absolute_uri, method, body, headers, redirect):
@@ -55,21 +54,21 @@ class Http(urllib3.PoolManager):
 
 
 class VerboseHttp(Http):
-    """A subclass of Http that verbosely reports on activity.
+    """A subclass of ``Http`` that verbosely reports on activity.
 
     If the output is a tty or ``GABBI_FORCE_COLOR`` is set in the
     environment, then output will be colorized according to ``COLORMAP``.
 
     Output can include request and response headers, request and
-    response body content (if of a printable content-type), or both.
+    response body content (if of a printable content type), or both.
 
     The color of the output has reasonable defaults. These may be overridden
     by setting the following environment variables
 
-    * GABBI_CAPTION_COLOR
-    * GABBI_HEADER_COLOR
-    * GABBI_REQUEST_COLOR
-    * GABBI_STATUS_COLOR
+    * ``GABBI_CAPTION_COLOR``
+    * ``GABBI_HEADER_COLOR``
+    * ``GABBI_REQUEST_COLOR``
+    * ``GABBI_STATUS_COLOR``
 
     to any of: BLACK RED GREEN YELLOW BLUE MAGENTA CYAN WHITE
     """
@@ -171,7 +170,7 @@ class VerboseHttp(Http):
 
 
 def get_http(verbose=False, caption=''):
-    """Return an Http class for making requests."""
+    """Return an ``Http`` class for making requests."""
     if not verbose:
         return Http(strict=True)
 


gabbi/httpclient.py
```diff
@@ -172,15 +172,10 @@ def _verbose_output(self, message, prefix='', color=None, stream=None):
 
 def get_http(verbose=False, caption=''):
     """Return an Http class for making requests."""
```
FND: `s/Http/HTTP/`?

gabbi/httpclient.py
```diff
@@ -172,15 +172,10 @@ def _verbose_output(self, message, prefix='', color=None, stream=None):
...
-        if verbose == 'headers':
-            body = False
-        return VerboseHttp(body=body, headers=headers, colorize=colorize,
-                           stream=stream, caption=caption, strict=True)
-    return Http(strict=True)
+    if not verbose:
+        return Http(strict=True)
+
+    headers = False if verbose == 'body' else True
+    body = False if verbose == 'headers' else True
```
FND: FWIW, this seems like a good (rare) example of ternaries actually being significantly more readable than the less concise alternative ("include headers/body unless verbosity is set to body/headers").

gabbi/httpclient.py
cdent: No, the class is named `Http`. It does `HTTP`.

gabbi/httpclient.py
cdent: 🤷‍♂️

gabbi/httpclient.py
FND: I've now added d47f9430f96e8be044d2097005186213c3a7d275 - feel free to omit that, as I'm not sure it's entirely correct.

gabbi/httpclient.py
```diff
@@ -10,7 +10,6 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-"""Subclass of Http class for verbosity."""
```
FND: This seemed wrong/outdated?

gabbi/httpclient.py
```diff
@@ -29,10 +28,10 @@
 
 
 class Http(urllib3.PoolManager):
-    """A subclass of the urllib3.PoolManager to munge the data.
+    """A subclass of the ``urllib3.PoolManager`` to munge the data.
 
     This transforms the response to look more like what httplib2
-    provided when it was used as the httpclient.
+    provided when it was used as the HTTP client.
```
FND: Not sure whether that's what's meant here or why it's relevant at all: As someone not intimately familiar with the code's evolution, that historical note is somewhat baffling, even confusing.

gabbi/httpclient.py
```diff
@@ -55,21 +54,21 @@ def request(self, absolute_uri, method, body, headers, redirect):
...
 class VerboseHttp(Http):
-    """A subclass of Http that verbosely reports on activity.
+    """A subclass of ``Http`` which verbosely reports on activity.
 
     If the output is a tty or ``GABBI_FORCE_COLOR`` is set in the
     environment, then output will be colorized according to ``COLORMAP``.
 
     Output can include request and response headers, request and
-    response body content (if of a printable content-type), or both.
+    response body content (if of a printable content type), or both.
```
FND: If you specifically mean the HTTP header, backticks and capitalization would help clarify that.

gabbi/httpclient.py
cdent: It is outdated, it was a subclass of httplib2's Http class. Now it is is a subclass of urllib3's PoolManager operating as an HTTP client (urllib3 does this layering thing where methods are cascaded up through various classes by a combination of composition and inheritance.

gabbi/httpclient.py
cdent: Something needs to be said there, otherwise, there's no reason for the transformation to be happening.

The transformation is happening because it used to be httplib2 that was the client and the gabbi uses those semantics internally.

So if you wanted to change this then incorporating some of that info could potentially be useful.

gabbi/httpclient.py
cdent: No, I mean the content type of the response, which will often be described by the header, but I don't mean the header itself. The hyphen was probably an incorrect(?) effort to try to mollify your concerns with my use of hyphens.

gabbi/httpclient.py
cdent: In addition the use of the term 'httpclient' was to draw an association with the file that you're currently editing, not the generic notion of an HTTP client. But either is fine.

gabbi/httpclient.py
```diff
@@ -55,21 +54,21 @@ def request(self, absolute_uri, method, body, headers, redirect):
 
 
 class VerboseHttp(Http):
-    """A subclass of Http that verbosely reports on activity.
+    """A subclass of ``Http`` which verbosely reports on activity.
```
cdent: That is correct here as it is required for the meaning of the sentence, see: http://www.writersdigest.com/online-editor/which-vs-that

gabbi/httpclient.py
```diff
@@ -55,21 +54,21 @@ def request(self, absolute_uri, method, body, headers, redirect):
...
-    * GABBI_CAPTION_COLOR
-    * GABBI_HEADER_COLOR
-    * GABBI_REQUEST_COLOR
-    * GABBI_STATUS_COLOR
+    * ``GABBI_CAPTION_COLOR``
+    * ``GABBI_HEADER_COLOR``
+    * ``GABBI_REQUEST_COLOR``
+    * ``GABBI_STATUS_COLOR``
 
     to any of: BLACK RED GREEN YELLOW BLUE MAGENTA CYAN WHITE
```
cdent: why backticks above but not here?

gabbi/httpclient.py
FND: heh My rule of thumb is: Hyphenate if it's a multi-word modifier (see?) - cf. [front end vs. front-end developer](https://css-tricks.com/poll-results-front-end-front-end/#article-header-id-0). Essentially, the hyphen acts like parentheses there, thus improving readabilty.

gabbi/httpclient.py
FND: Glad you noticed. I figured that'd be overkill here, doing more harm than good.

gabbi/httpclient.py
FND: Off the top of my head, I'm not sure how to improve this, so might as well not worry about it for the moment.

gabbi/httpclient.py
FND: Huh, I'd learned pretty much the opposite - might just be a different school of thought, I don't have a reference handy right now. Reverted back to "that" in f13aa5ee0bc6e5d73f3828126f03cc60a4c5254f.

gabbi/httpclient.py
FND: You might not care, but I'll be sure to weaponize this in future discussions.

cdent:
I forget where we left this in our discussions. Go? No go?

FND:
Python, I believe.

Looks to me like we had a thorough discussion, the change still seems valuable to me. Go.

PR #243, opened 2018-04-13
Allow listing of tests with no host configured

zaneb:
When creating tests, if the host was empty (and no intercept was provided),
an AssertionError would be raised immediately. This meant that in the case
where no configuration information is available, authors of load_test()
functions had to choose between simply not loading any tests (or
suppressing the exception) and allowing test discovery to fail.

This change allows an empty string to be passed as the host, in which case
the TestCases will be generated (so tests can be listed) but the tests will
be skipped. This will show up in reporting, whereas excluding the tests
from the listing is silent.

If the host, url, and intercept are all None, this will continue to
generate an immediate AssertionError since it is likely due to the caller
simply not passing any of those arguments.

See https://bugs.launchpad.net/heat-tempest-plugin/+bug/1749218 for the initial request.

### DIFF
--- a/gabbi/driver.py
+++ b/gabbi/driver.py
@@ -83,7 +83,7 @@ def build_tests(path, loader, host=None, port=8001, intercept=None,
 
     # Exit immediately if we have no host to access, either via a real host
     # or an intercept.
-    if not bool(host) ^ bool(intercept):
+    if not ((host is not None) ^ bool(intercept)):
         raise AssertionError(
             'must specify exactly one of host or url, or intercept')
 
--- a/gabbi/suitemaker.py
+++ b/gabbi/suitemaker.py
@@ -106,8 +106,10 @@ class TestMaker(object):
         # from a different module.
         if self.test_loader_name:
             klass.__module__ = self.test_loader_name
+        check_host = case.testcase.skipIf(not self.host,
+                                          'No host configured')
 
-        tests = self.loader.loadTestsFromTestCase(klass)
+        tests = self.loader.loadTestsFromTestCase(check_host(klass))
         history[test['name']] = tests._tests[0]
         # Return the first (and only) test in the klass.
         return tests._tests[0]


zaneb:
@cdent not gonna die in a ditch over this or anything, but interested to see what you think about the right way to handle. (Maybe we should fail the tests instead of skipping?)

cdent:
Interesting. I'm okay with this happening. Input from @FND,  @jasonamyers  and @jd would be useful.

I _think_, but am not certain, that failing would be better than skipping. What's your mental calculus on why you went with skipping?

cdent:
In general gabbi tends to want to be fail fast because I think of it as a tools used by humans and this human (me) wants to know things are wrong as early as possible. 

In other situations, though, I can see how that would be a problem.

zaneb:
I guess my mental calculus was that that's [what we're doing right now](https://review.openstack.org/422710), but we're doing it silently and I'd at least like it to be reported. I guess you can always make it fail by passing garbage, but there's no way to make it skip at the moment.

My main issue with our [current proposal](https://review.openstack.org/544025) though is that it will fail in bizzarro ways. So if it failed at runtime with a sensible error message I think that would be OK too.

gabbi/driver.py
```diff
@@ -83,7 +83,7 @@ def build_tests(path, loader, host=None, port=8001, intercept=None,
 
     # Exit immediately if we have no host to access, either via a real host
     # or an intercept.
-    if not bool(host) ^ bool(intercept):
+    if not ((host is not None) ^ bool(intercept)):
```
FND: FWIW, this seems easier to grok:

```python
if (host is None) ^ (bool(intercept) is False):
```

(or some variation thereof)

I had to look up `^` meaning XOR, but not sure whether that could be simplified.

jd:
Looks fine with me.
There's always a danger that you forget to set a host and then the test are skipped and you don't realize, but at least this time they will be reported as skipped, whereas the workaround to implement in every project will anyway do nothing silently.
That sounds like a good tradeoff.

cdent:
This has introduced some issues in the way telemetry is using gabbi. They intentionally have no host nor intercept set when they call test_suite_from_dict because all of their tests have a fully qualified url. The workaround is to set a random host that gets ignored.

But it seems we can probably do a bit better. IRC log at http://p.anticdent.org/1Umv @zaneb any ideas?

zaneb:
Fixed in #244

PR #244, opened 2018-04-19
Fix issues with skipping tests when unconfigured

zaneb:
Mea culpa.

This will fix things while we wait for https://github.com/testing-cabal/testtools/pull/268


cdent:
@sileht can you try this in the telemetry set up?

cdent:
Yay! Thanks @sileht . Thanks @zaneb

PR #251, opened 2018-06-22
read data from disk improvement

### LINKED ISSUE #250: jsonhandler: read data from disk improvements
UPDATE: dropping the yaml loading from this issue and updating the querying of the subset of data from disk to enable JSONPath support like the left hand side does already for the jsonhandler. For example, $.versions[?status = "current"].version: <@endpoint.json:[?status = "current"].version

reading data from disk can be done today with the jsonhandler by using the '<@myfile.json' we would like to improve this to work on larger files by allowing a user to specify a subset. Also being able to read from a yaml source would be nice.

1. allow yaml to be loaded from disk this should be easy since json is supposed to be a subset of it. 
    - a foreseen problem I see with this is the test loader grabs all *.yaml files, we can get around this without modifying the test loader by naming our test yaml files with a .yml extension

2. Allow a user to specify a subset to load. It appears to be a simple change. I think we can use the ‘:’ character (or whatever we agree on) to split on the accessors for the dictionary and apply those to the self.loads(info) call like so https://github.com/trevormccasland/gabbi/commit/5a17517bb52fe6d4206cee42e63bca9d73ee8f35 ( I forgot to swap out the json.loads with yaml, ignore the point in the commit message :/ )
    - Our problem is the management of the json files, we would prefer that we could load our site deployment information in a yaml format and use those values for some of our api responses without having to create a json file for every test.

The following is a simple example, the contents of the files for our use case will be much larger and contains information that is related to how our apis are deployed and do not necessarily correlate to a single api response.

Example:
  endpoint.json contents:
    '[{"status": "current", "version": 1.0}]'

  response_json_paths:
    $.versions[0][‘status’]: <@endpoint.json:[0][status]

trevormccasland:
This commit aims to add a new feature that allows a user to specify
a subset of a json file by using the existing JSONPath functionality
by expanding on the read from disk operator '<@'

closes #250

### DIFF
--- a/docs/source/jsonpath.rst
+++ b/docs/source/jsonpath.rst
@@ -78,6 +78,42 @@ or::
 
 Examples like this can be found in one of gabbi's `own tests`_.
 
+When reading from disk you can apply the same JSONPath by adding a ':' to the
+end of your file name. This allows you to store multiple api responses into
+a json file to reduce file management when constructing your tests
+
+.. highlight:: json
+
+Given JSON data as follows::
+
+    {
+        "values": [{
+            "pets": [{
+                "type": "cat",
+                "sound": "meow"
+            }, {
+                "type": "dog",
+                "sound": "woof"
+            }]
+        }, {
+            "people": [{
+                "name": "chris",
+                "id": 1
+            }, {
+                "name": "justin",
+                "id": 2
+            }]
+        }]
+    }
+
+.. highlight:: yaml
+
+You can write your tests like the following::
+
+    response_json_paths:
+        $.pets: <@pets.json
+        $.pets[?type = "cat"].sound: <@values.json:$.values[0].pets[?type = "cat"].sound
+
 There are more JSONPath examples in :doc:`example` and in the
 `jsonpath_rw`_ and `jsonpath_rw_ext`_ documentation.
 
--- a/docs/source/jsonpath.rst
+++ b/docs/source/jsonpath.rst
@@ -79,8 +79,8 @@ or::
 Examples like this can be found in one of gabbi's `own tests`_.
 
 When reading from disk you can apply the same JSONPath by adding a ':' to the
-end of your file name. This allows you to store multiple api responses into
-a json file to reduce file management when constructing your tests
+end of your file name. This allows you to store multiple API responses into
+a JSON file to reduce file management when constructing your tests.
 
 .. highlight:: json
 
@@ -113,6 +113,10 @@ You can write your tests like the following::
     response_json_paths:
         $.pets: <@pets.json
         $.pets[?type = "cat"].sound: <@values.json:$.values[0].pets[?type = "cat"].sound
+
+Although placing more than one API response into a single JSON file may seem
+convenient, keep in mind there is a tradeoff in readability that should not
+be overlooked before implementing this technique.
 
 There are more JSONPath examples in :doc:`example` and in the
 `jsonpath_rw`_ and `jsonpath_rw_ext`_ documentation.
--- a/gabbi/handlers/jsonhandler.py
+++ b/gabbi/handlers/jsonhandler.py
@@ -76,29 +76,45 @@ class JSONHandler(base.ContentHandler):
     def action(self, test, path, value=None):
         """Test json_paths against json data."""
         # Do template expansion in the left hand side.
-        path = test.replace_template(path)
+        lhs_path = test.replace_template(path)
+        rhs_path = rhs_match = None
         try:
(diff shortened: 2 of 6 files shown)


trevormccasland:
I think a slight improvement can be made on the split operator by providing the second argument as a 1. This will make keys that include the ':' character safe to use in the program. When I update it, should I provide a test too?

gabbi/handlers/jsonhandler.py
```diff
@@ -76,29 +76,45 @@ def extract_json_path_value(data, path):
...
-            raise AssertionError('json path %s cannot match %s' %
-                                 (path, test.response_data))
+            raise AssertionError('left hand side json path %s cannot match '
+                                 '%s' % (path, test.response_data))
 
         # read data from disk if the value starts with '<@'
         if isinstance(value, str) and value.startswith('<@'):
+            # Do template expansion in the rhs if rhs_path is provided.
+            if ':' in value:
+                value, rhs_path = value.split(':')
```
cdent: We probably need to think about how we want to manage the separator. If the filename or the jsonpath has a `:` in it, we've got some issues. Unless we want to do tokenizing (please no!) there's not going to be a foolproof way to deal with that, but there are probably some ways we can make this more robust, but I'm not sure which is the best. Some ideas:

* if we assume there's no `:` in the jsonpath we can rsplit
* if we assume there's `:` in the file we can split, 1 as you've mentioned elsewhere
* we could split on `:$` (thinking if it as a signal for "here comes some jsonpath)
* we could split on `::` or ` :: ` as mentioned in the IRC chat

I don't have a strong opinion on this, but it is worth taking the time to think up something.

docs/source/jsonpath.rst
```diff
@@ -78,6 +78,42 @@ or::
 
 Examples like this can be found in one of gabbi's `own tests`_.
 
+When reading from disk you can apply the same JSONPath by adding a ':' to the
+end of your file name. This allows you to store multiple api responses into
+a json file to reduce file management when constructing your tests
```
cdent: Need a '.' on the end of the sentence?

Also I think it would be wise if we had a small warning here about the impact that putting results in a different file that has to be parsed may have on the readability of the tests. Just a light warning, not something that would dissuade people from doing it, but just to make them aware of the cost/benefit balance.

docs/source/jsonpath.rst
```diff
@@ -78,6 +78,42 @@ or::
 
 Examples like this can be found in one of gabbi's `own tests`_.
 
+When reading from disk you can apply the same JSONPath by adding a ':' to the
+end of your file name. This allows you to store multiple api responses into
```
FND: Small detail: For readability and consistency, please capitalize abbreviations like "API" and "JSON" (below).

trevormccasland:
I definitely have the time, I was planning on submitting another issue / pull request for the yaml-based test.

cdent:
> I definitely have the time, I was planning on submitting another issue / pull request for the yaml-based test.

Just to be clear I mean a yaml file in `gabbi/tests/gabbits_intercept/` that uses the syntax developedin this patch.

trevormccasland:
Yeah that does make it clear! I'll add a yaml test their too.

gabbi/tests/test_unsafe_yaml.py
```diff
@@ -33,6 +33,9 @@
 
 
 yaml.add_constructor(u'!IsNAN', lambda loader, node: util.NanChecker())
+yaml.add_constructor(
```
cdent: I'll fix this on master, as it is breaking there too. It's because of the new release of yaml. I'll need to update the requirements to make everything smooth.

cdent: Turns out this is considerably more complicated than I thought. #252 is in progress to try to fix it.

trevormccasland: Oh that makes sense why I couldn't reproduce the error. I'll wait for #252 before making a change to remove my last commit one (56c9471).

cdent: I've changed master to not use PyYAML above >=4.0 for the time being, 4.1 is probably going to change.

If you remove your last patch, rebase onto master (or merge from it), I think we'll be good to go
and we can release this change and the yaml guard. Once I figure out what's really going on with 4.x we can look back at #252

cdent:
Thanks for that @twm2016. I've gone ahead and merged it. I'm still mulling over how best to handle the yaml issues, but mostly likely will do a release later today with PyYAML constrained, and with this new feature. Thanks very much.

PR #252, opened 2018-06-28
Fix object handling in yaml for PyYAML 4.x

cdent:
PyYAML 4.x is "safe" by default and in that context safe
means that arbitrary python objects are not okay. However,
defined tags which are added by the running process are
safe.

This change updates gabbi so that if unsafe YAML is desired,
the new "danger_load" function is used and expands the
NanChecker related tests to reflect this different
understanding of safe and unsafe. I suspect our tests
were incorrect in the past.


cdent:
This is covered by #267

PR #254, opened 2018-06-29
jsonhandler: allow reading yaml data from disk

### LINKED ISSUE #253: jsonhandler: allow reading yaml data from disk
reading data from disk can be done today with the jsonhandler by using the '<@myfile.json' notation we would like to improve this to work on yaml files by allowing a user to specify a file in the test's sub directory like so, '<@subdir/myfile.yaml'

trevormccasland:
This commit aims to change the jsonhandler to be able to read
data from disk if it is a yaml file.

Note:
  * Simply replacing the loads call with yaml.safe_load is not enough
    due to the nature of the NaN checker requiring an unsafe load[1].

closes #253
[1] https://github.com/cdent/gabbi/commit/98adca65e05b7de4f1ab2bf90ab0521af5030f35


docs/source/jsonpath.rst
```diff
@@ -76,11 +76,17 @@ or::
         $.pets: <@pets.json
         $.pets[0]: <@cat.json
 
+to use YAML files like the JSON ones above, they must be placed in a
+subdirectory::
```
cdent: may as well explain why

gabbi/handlers/jsonhandler.py
```diff
@@ -95,7 +96,10 @@ def action(self, test, path, value=None):
                 rhs_path = test.replace_template('$' + rhs_path)
             info = test.load_data_file(value.replace('<@', '', 1))
             info = six.text_type(info, 'UTF-8')
-            value = self.loads(info)
+            if value.endswith('yaml'):
+                value = yaml.safe_load(info)
```
cdent: You say in the commit message that you can't use solely `yaml.self_load()` because of the NanChecker, but it is not clear how that is related to this.

trevormccasland: Yeah good call, Looking at this again, I believe this has to do with a bug on how we test for isnan in tests/util.py we should cast the local var 'other' as a float because NaN is an IEEE 754 special value
(Pdb) other
'NaN'
(Pdb) math.isnan(other)
*** TypeError: a float is required

docs/source/jsonpath.rst
trevormccasland: will do.

trevormccasland:
Next patch will have the new ContentHandler subclassing JSONHandler and overriding the loads method and hopefully a better fix for the NaN issue. Do we want to fail if the input is a string? I believe the gnocchi tests are telling me yes, so maybe its not an issue afterall.

trevormccasland:
So I think I see why the test case is failing now, it appears a float('NaN') dumped by json converts to a string when loaded in yaml.
>>> myd = {'a': float('NaN')}
>>> j = json.dumps(myd)
>>> y = yaml.dump(myd)
>>> y
'{a: .nan}\n'
>>> j
'{"a": NaN}'
>>> yaml.__version__
'3.12'
>>> yaml.safe_load(j)
{'a': 'NaN'}
>>> yaml.safe_load(y)
{'a': nan}

I'm still working on the YAMLHandler. I'm refactoring the ContentHandler base class to provide a load_data_file method that returns python data to get around this issue.

cdent:
Yeah, it's pretty clear that despite the rumors, at least in Python, JSON is not a strict subset of YAML and you can't load it with complete accuracy. I think this makes it a pretty clear that using as the loader needs to be something that is opt-in, and the way we do that with ContentHandlers is by having others.

I agree that having an overridable `load_<something>` method of some kind is a good way to go.

Will look through the current patches to see what else is up. Thanks for plugging away at this.

gabbi/handlers/jsonhandler.py
```diff
@@ -54,7 +55,7 @@ def dumps(data, pretty=False, test=None):
 
     @staticmethod
     def loads(data):
-        return json.loads(data)
+        return yaml.safe_load(data)
```
cdent: The root of the speedbumps you've hit is because you're engaging `yaml.safe_load` here, instead of in the `action` method, meaning yaml gets involved in all the JSON processing, not just when you are loading data from disk to be handled in a `response_json_paths`.

This might be the right thing, but did you consider doing it in `action` and keeping the behaviors separated?

docs/source/handlers.rst
```diff
@@ -22,6 +22,12 @@ string into a data structure in ``response_data`` that is used when
 evaluating ``response_json_paths`` entries in a test or doing
 JSONPath-based ``$RESPONSE[]`` substitutions.
 
+A YAML content handler has been added to extend the JSON handler.
```
cdent: We need to call it something different than the YAML content handler because what the name `YAMLContentHandler` means is that this is a handler for request and/or response bodies which are YAML. That's not what this is.

Naming is hard, so I don't have any clever ideas at the moment but will think on it.

gabbi/driver.py
```diff
@@ -105,7 +105,6 @@ def build_tests(path, loader, host=None, port=8001, intercept=None,
     for handler in (content_handlers + response_handlers +
                     handlers.RESPONSE_HANDLERS):
         handler_objects.append(handler())
-
```
cdent: unrelated?

gabbi/case.py
```diff
@@ -190,9 +190,12 @@ def load_data_file(self, filename):
 
     def _assert_response(self):
         """Compare the response with expected data."""
+        handlers = self.response_handlers
+        content_handler = self.get_content_handler(self.content_type)
+        if content_handler:
+            handlers.append(content_handler)
```
cdent: This isn't the right fix for this. Unfortunately, we're going to need to dig a bit to figure out what the right fix is. The problem will be happening somewhere in the `_register` handling in the base `ResponseHandler`.

Part of the reason this is wrong is that this method is supposed to be using response handlers only (of which content handlers are a subclass, but not all responsehandlers are content handlers) so if you have to go looking in get_content_handler than a handler has not _registered properly, or has been clobbered somehow.

I haven't got the cycles to dig into it today.

gabbi/handlers/base.py
```diff
@@ -12,7 +12,6 @@
 # under the License.
 """Base classes for response and content handlers."""
 
-
```
cdent: unreleased (although this is a fine tidy up)

gabbi/suitemaker.py
```diff
@@ -219,10 +219,10 @@ def test_suite_from_dict(loader, test_base_name, suite_dict, test_directory,
     default_test_dict = copy.deepcopy(case.HTTPTestCase.base_test)
     for handler in handlers:
         default_test_dict.update(handler.test_base)
-        if handler.response_handler:
-            response_handlers.append(handler.response_handler)
         if handler.content_handler:
             content_handlers.append(handler.content_handler)
+        elif handler.response_handler:
+            response_handlers.append(handler.response_handler)
```
cdent: again this is probably not the right fix, as when a handler registers it can declare itself as both a content handler and a response handler and they are not mutually exclusive. For some things (notably the JSONHandler) it is both.

trevormccasland: I added a class attribute so a handler can declare itself as just a content handler.

gabbi/case.py
trevormccasland: I added a class attribute to help with this.

docs/source/handlers.rst
trevormccasland: I went with the ugly but explicit route :/ YAMLDiskLoadingJSONHandler . Alternatives could be YDLJSONHandler or YDLJHandler for short? Any other suggestions are welcome too.

docs/source/handlers.rst
```diff
@@ -149,4 +153,4 @@ passed the ``response_data`` of the prior test and the argument within the
 Please see the `JSONHandler source`_ for additional detail.
 
 .. _JSONHandler source: https://github.com/cdent/gabbi/blob/master/gabbi/handlers/jsonhandler.py
-.. _YAMLHandler tests: https://github.com/cdent/gabbi/blob/master/gabbi/tests/test_yaml_handler.py
+.. _YAMLDiskLoadingJSONHandler tests: https://github.com/cdent/gabbi/blob/master/gabbi/tests/test_yaml_disk_loading_jsonhandler.py
```
trevormccasland: I don't see this being used anywhere :/

docs/source/jsonpath.rst
```diff
@@ -76,14 +76,20 @@ or::
...
+   runner from consuming them as test files to run::
 
     response_json_paths:
         $: @<subdir/values.yaml
 
-Examples like this can be found in one of gabbi's `own tests`_.
-
 When reading from disk you can apply the same JSONPath by adding a ':' to the
 end of your file name. This allows you to store multiple API responses into
 a JSON or YAML file to reduce file management when constructing your tests.
```
trevormccasland: this may just have to read as '...multiple API responses into a JSON file...'

cdent:
Writing just to ack that I've had a look at this, but will need to look more closely tomorrow to more fully understand the added class attribute.

Also, whatever is going on with the failing pytest-driven tests needs to be addressed.

Thanks for plugging away.

cdent:
I'm going to try some experiments myself. I think one of the issues here is that the new subclassed handler should replace the json handler entirely, but I'm struggling to trace the logic in my brain so I need to hack on the code. I'll post back here when I've had some more concrete thoughts.

cdent:
That need for different named flags is almost certainly a bug in the pytest handling. Good find. I'll try to fix that.

gabbi/suitemaker.py
trevormccasland: I noticed this change hasn't been reverted yet. Looks like there is still some more work to do on this part.

cdent: I've got a branch in progress that tries to address that in a slightly different way. We'll want to compare notes on that and decide which is best. I'll push it up shortly.

trevormccasland: ok I'll wait on that then. I found there was a typo in jsonhandler.py too that this was covering up.
-    add_reponse_handling = False
+    add_response_handling = False

gabbi/suitemaker.py
```diff
@@ -217,12 +217,16 @@ def test_suite_from_dict(loader, test_base_name, suite_dict, test_directory,
 
     # Merge global with per-suite defaults
     default_test_dict = copy.deepcopy(case.HTTPTestCase.base_test)
+    seen_keys = set()
     for handler in handlers:
         default_test_dict.update(handler.test_base)
+        if handler.response_handler:
+            key = 'response_%s' % handler.test_key_suffix
```
cdent: since key is getting thrown away outside this look, constructing 'key' from 'response_%s' is probably not necessary (as it will be done for every suffix). Should be okay to use `handler.test_key_suffix` directly in the set.

cdent:
This is looking good. I adding one more comment on a small thing that could be fixed to make it a bit more futureproof.

Thanks!

cdent:
This is release as 1.44.0

@trevormccasland If you get the opportunity to write a blog post or something like that about how you are using gabbi, especially the features you added, that would be great. Since people rarely read the docs all the way through, only something that does a bit of "we wanted to do this, so we added these features" kind of explanation will really get the idea across.

Thanks again.

trevormccasland:
Thanks a lot for the reviews and working through the pull request with me. I will work on a blog post as soon as we get everything together on our side. Then I'll post the link here when it's ready.

PR #256, opened 2018-07-09
WIP: alternate mode for repeating response handlers

cdent:
This is an alternate version of pull #254, with the difference
in gabbi/suitemaker.py where response_handlers is only extended
if the response_* key has not been seen before. This means that
if the YAMLDiskLoadingJSONHandler is turned on it does all the
work of the JSONHandler for respone handling, because it is only
present once.

At the same time, the existing handing for content handlers means
that custom handlers which accept a content-type are used first (and
only) for the content type it accepts.

The issues with pytest and files having the same name remain
here. I tried to make some change to pytest test loading to
make the global cache use more distinguishing keys, but it was
distracting from progressing on the other work, so I'll deal with
that separately.


cdent:
We don't necessarily want to merge this (in part because it uses a private member), but it provides a demo of a potentially cleaner way to deal with duplicated response handlers.

@trevormccasland what do you think?

cdent:
#253 fixed this, so closing

PR #260, opened 2018-10-04
Tweak intro to be more concise and welcoming

FND:
this is a work in progress; a quick first attempt to provide some food for thought

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -25,24 +25,24 @@ Gabbi
 .. highlight:: yaml
 
 Gabbi is a tool for running HTTP tests where requests and responses
-are expressed as declarations in a collection of YAML files. The
-simplest test looks like this::
+are expressed as declarations in YAML files::
 
     tests:
-    - name: A test
-      GET: /api/resources/id
+    - name: retrieve items
+      GET: /items
 
 See the rest of these docs for more details on the many features and
 formats for setting request headers and bodies and evaluating responses.
+
+Tests can be run from the command line with :doc:`gabbi-run <runner>` or
+programmatically using either py.test or
+:ref:`unittest <test_loaders>`-style test runners.
 
 The name is derived from "gabby": excessively talkative. In a test
 environment having visibility of what a test is actually doing is a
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -25,25 +25,25 @@ Gabbi
...
-      GET: /api/resources/id
+    - name: retrieve items
+      GET: /items
 
 See the rest of these docs for more details on the many features and
 formats for setting request headers and bodies and evaluating responses.
 
+Tests can be run from the command line with :doc:`gabbi-run <runner>` or
+programmatically using either py.test or
+:ref:`unittest <test_loaders>`-style test runners.
```
FND: In my experience, installation instructions are one of the biggest hurdle in getting people started - which is why I'd devoted an entire section to this in https://www.innoq.com/en/blog/gabbi-declarative-http-testing/.

Not quite sure how to weave that into this page though...

cdent: there's a disjunction between what's in the readme and what's in the front page of the docs that could probably be usefully resolved

FND: I understood the README to target contributors rather than users - but perhaps I'm misunderstanding what you have in mind?

cdent: just that the README is more direct about installation

FND: Oh, right - how about this then:

> ### Installation
>
> As a Python package, gabbi is typically installed via pip:
> ```
> pip install gabbi
> ```
> You might want to create a virtual environment (an isolated Python context): <instructions from the aforementioned article's info box>

cdent: sounds good to me

FND: I've added 1215cd774d231f316c552febcce2a5e24b61f435 here; let me know if you want me to merge that myself.

(Note that I'm pretty ignorant WRT ReST though.)

cdent: It'll need some tweaks, I'll get them.

FND: Ah, excellent - thank you!

PR #264, opened 2018-12-06
Update testing to use python 3.7

cdent:
Python 3.7 is common now, so make it the main test version
in tox.ini and update .travis.yml accordingly.

To have 3.7 we need to start using the xenial dist in travis, and
fall back to trusty for pypy.

In python 3.7 re.escape changes from escaping all non-alphanumeric,
to just special characters, requiring a change in test_history
for that version.

Trove classifiers are updated to reflect 3.7 support. Support for
3.4 is removed, not because gabbi doesn't work on 3.4 but because
it is not worth indicating support.

### DIFF
--- a/gabbi/tests/test_history.py
+++ b/gabbi/tests/test_history.py
@@ -13,6 +13,7 @@
 """Test History Replacer.
 """
 
+import sys
 import unittest
 
 from gabbi import case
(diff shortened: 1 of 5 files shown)


gabbi/tests/test_history.py
```diff
@@ -120,7 +121,10 @@ def test_cookie_replace_prior_regex(self):
 
         cookie = self.test('test_request').replace_template(
             self.test.test_data, escape_regex=True)
-        self.assertEqual(r'/test\=cookie\?/', cookie)
+        if sys.version_info[:2] >= (3, 7):
+            self.assertEqual('/test=cookie\?/', cookie)
+        else:
+            self.assertEqual('/test\=cookie\?/', cookie)
```
FND: You might wanna turn that bit from the commit message into a comment here - though I'm not actually convinced it's worth the bloat.

cdent: Yeah, I waffled on that. This makes it pretty clear that _something_ changed in 3.7 but not what and I'd probably be annoyed by needing to git blame to find out what it was.

But then again, annoying people can be fun.

PR #271, opened 2019-07-01
Adapt to introduced by release of pytest5

cdent:
pytest 5 currently does not work with the gabbi tests because of:

* deprecation warnings due to `pytest.config` no longer being preferred
* support for python 2 being dropped

While that is resolved, release a version that is `pytest<5.0.0`


cdent:
The test failures are in placement, unrelated to gabbi, and being fixed simultaneously, so not going to wait.

PR #274, opened 2019-07-30
add cert validation attribute (runner + driver)

TheMatrix97:


cdent:
Awesome, thanks for doing this, will give it a quick review today and then get some other people involved too.

gabbi/driver.py
```diff
@@ -132,6 +133,12 @@ def build_tests(path, loader, host=None, port=8001, intercept=None,
             else:
                 suite_dict['defaults'] = {'verbose': verbose}
 
+        if any((cert_validate == opt for opt in [True, False])):
```
cdent: Since `cert_validate` is a boolean and we want any arg type issues to bubble up quickly when someone is building their test system, and since the default is already set in the signature of the containing method (`build_tests`) I think you can model the code here on what's at line 123 and used by `require_ssl`:

```python
if not cert_validate:
    if 'defaults' in suite_dict:
       suite_dict['defaults']['cert_validate'] = cert_validate
    else:
        suite_dict['defaults'] = {'cert_validate': cert_validate}
```

or something like that. Similar comments apply in other places that you've used the same idiom.

gabbi/runner.py
```diff
@@ -241,6 +249,13 @@ def _make_argparser():
         choices=['all', 'body', 'headers'],
         help='Turn on test verbosity for all tests run in this session.'
     )
+    parser.add_argument(
+        '--no-cert-validate',
+        dest='cert_validate',
+        action='store_false',
+        default=True,
```
cdent: argparse is confusing...

cdent:
Thanks for the updates. I should have a chance to look properly tomorrow.

gabbi/runner.py
```diff
@@ -136,11 +136,11 @@ def run_suite(handle, handler_objects, host, port, prefix, force_ssl=False,
             data['defaults']['verbose'] = verbosity
         else:
             data['defaults'] = {'verbose': verbosity}
-    if cert_validate is not None and not cert_validate:
+    if not cert_validate:
```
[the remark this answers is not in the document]
cdent: correct

gabbi/runner.py
cdent: Sorry, my comment was unclear. What I was saying was that the combo of:

```
action='store_false',
default=True,
```

was weird.

I have no strong opinion on `--ignore-cert` v `--no-cert-validate`, whichever you prefer.

cdent:
> I got stuck writting a test case for this feature... I cannot find a way to get the Http Object and somehow check if the cert validation is on or off...

If you want to leave that part, I can do it once we are happy with the rest of it.

cdent:
Let me know how you'd like to proceed with the `gabbi-run` argument. If you're happy as is, we can go ahead and get this merged and I'll add a test for the arg confirmation.

cdent:
I'm going to go ahead and merge this myself and make some adjustments to add tests and the like. Thanks very much for the contribution. Once I have things tweaked to my liking I'll make a new release.

TheMatrix97:
Hey Chris,
Sorry for the delay, actually I'm on a trip :D
I copy pasted the parser add argument from unsafe-yaml... I agree it could
be a bit of boolean confusion, but for me it was the easiest way to proceed.
If you need something heading the next release please contact me :)

cdent:
1.49.0 was just released with this functionality to https://pypi.org/project/gabbi/

the main difference is added docs and the args are `-k` and `--insecure` to mirror `curl`.

And after all that I forgot to add tests... 🤷‍♂

TheMatrix97:
Wow that was fast!
Thank you! :)

Issue #270, opened 2019-06-19
gabbi yaml does't support passing string "number" which defined in bash environment


taget:
pip freeze | grep gab
gabbi==1.46.0

1. environment variable
(venv) [root@VM_62_220_tlinux /data/eli/vs_tests]# cat source.rc | grep OWNER
export OWNER='251006071'

2. yaml file
(venv) [root@VM_62_220_tlinux /data/eli/vs_tests]# cat create_instance.yaml | grep OWNER
            owner: "$ENVIRON['OWNER']"

3. but when I run this yaml with gabbit
I got this
          "owner": 251006071,

expect "owner": "251006071",

================================

how I run gabbi:

[root@VM_62_220_tlinux /data/eli/vs_tests]# cat create_instance.yaml | grep OWN
            owner: "$ENVIRON['OWNER']"

ZONE_ID=100002 USER_ID='280078820' OWNER='251006071' [other environment] gabbi-run  http://10.104.62.34/ < create_instance.yaml

got result:
          "owner": 251006071,

expect:
          "owner": "251006071",

cdent:
The reason this is happening is because of the code in `_environ_replace`: https://github.com/cdent/gabbi/blob/master/gabbi/case.py#L217-L241

which casts things that look like numbers to numbers after the the yaml and environ processing.

I'm doing some experiments to see if there's a way to work around this and I'll post back.

cdent:
e3fb2cfa398e5591ba is the commit where this functionality was done.

The history was that someone had the opposite problem as you: anything that is in an environment variable is always a string and there was no clear way to make a number be a number.

The challenge is that the replace functionality is after the yaml has been processed, this is different from how things like ansible work, which does pre-processing of the template.

Fixing this is going to take some thought, I don't have any immediate ideas. Basically there needs to be a way to work around the casting.

I'll keep thinking about it. If you have any ideas, please respond with them.

cdent:
After thinking about it more, there's no easy way to address this without changing gabbi, and because of the existing precedent, it's hard to change things without break existing tests that are out there in the world.

However, one option for a potential fix is to put type "hints" in the name of the environment variable and then interpret them. Something like ['FOOBARint'] (but better than that).

Needs further thought. @taget would that potentially work for you?

taget:
@cdent 
[...]

PR #275, opened 2019-08-14
270 replacer casts

cdent:
This is an effort to provide the functionality described in #270 by implementing optional "casts" on the $ENVIRON and $REPLACER replacers by appending a :cast (int, str, bool, float) to the replacer name (e.g., `$ENVIRON:int['FOOBAR']`).

This only works when the replacer is the entirety of the message being replaced:

```yaml
data:
    foo: $ENVIRON:int['FOOBAR']
```

not mixed in:

```yaml
data:
    foo: boom $ENVIRON:int['FOOBAR'] boom
```

We can't do this for two reasons:

* in part, the surrounding context may conflict with the intended value of the cast
* but more, the way the `repl` function in`re.sub` works, they can only return strings, thus the casting has to happen after the re.sub

While the code is present to make it possible to do casts of other replacers, for the moment any cast will be ignored, mostly because it doesn't make sense for most of them.

It _might_, however, make sense for `$COOKIE` so input on that desired.

This isn't done, it needs to be confirmed as having grammatical sense, and it also needs doc update, but I'm putting it up here for some initial feedback on the format and implementation.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -406,8 +406,11 @@ class HTTPTestCase(unittest.TestCase):
     def _replacer_regex(key):
         """Compose a regular expression for test template variables."""
         case = HTTPTestCase._history_regex
-        return r"%s\$%s\[(?P<quote>['\"])(?P<arg>.+?)(?P=quote)\]" % (
-            case, key)
+        return (
+            r"%s\$%s(:(?P<cast>\w+))?"
+            r"\[(?P<quote>['\"])(?P<arg>.+?)(?P=quote)\]" % (
+            case, key))
+
 
     @staticmethod
     def _simple_replacer_regex(key):
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -52,6 +52,14 @@ REPLACERS = [
     'RESPONSE',
 ]
 
+# The list of approved type casts and associated functions
+APPROVED_CASTS = {
+    'int': int,
+    'float': float,
+    'str': str,
+    'bool': bool
+}
+
 # Basic test template determining both valid keys and default values
 BASE_TEST = {
     'name': '',
@@ -140,6 +148,7 @@ class HTTPTestCase(unittest.TestCase):
 
         If there is a prior test in the sequence, run it first.
         """
+        self.cast = None
         if self.has_run:
             return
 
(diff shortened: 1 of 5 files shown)


cdent:
@jd I suspect you might have some input on this if you can cast (ha!) your mind back to gnocchi time.

jd:
I think we mostly used `ENVIRON` in headers not in data, so we were cool with strings only.

Now I understand why you'd need this feature in cast.

gabbi/case.py
```diff
@@ -225,29 +245,33 @@ def _environ_replace(self, message, escape_regex=False):
...
+                    value = int(value)
+                return value
+            except ValueError:
+                pass
+            if value.lower() == "false":
+                return False
+            if value.lower() == "true":
+                return True
+            if value.lower() == "null":
+                return None
```
FND: minor nit:
```python
return {
    "true": True,
    "false": False,
    "null": None
}.get(value.lower(), value)
```
seems more readable/idiomatic?

cdent: Perhaps, but code was already there in the previous way and changing it now would be an unrelated change.

Also that looks icky.

cdent:
@taget have you had a chance to try this out?

cdent:
@taget I'm still hoping you can verify that this solves the problems you were trying to solve

taget:
@cdent

taget:
@cdent 
hi Chris, sorry to not response you for long time.

I tried in my test environment, it seems no different even I use the casts

yaml:

`
  POST: /api/cgi
  request_headers:
    content-type: application/json
  data:
    params:
      content:
        requestInfo:
          requestKey: mx_ksdionuj3djfk
          caller: vs-api
          requestModule: csig
          seqId: ''
          requestId: ''
          operator: csig
        version: '1.0'
        type: json
        requestItem:
          data:
            detail:
            - uuid: $ENVIRON['VS_UUID']
              hostIpList:
                  - $ENVIRON['VS_HOSTIP']
              migrateBandwidth: 20
              migrateTimeout: 3600
              pauseTime: $ENVIRON:int['PAUSE_TIME']
              schedulerDebug: 'on'
#              rawRootSwitch: 'on'
              compression:
                compression: mt
          method: instance_live_migrate
`

VS_HOSTIP="101" VS_UUID="100" PAUSE_TIME=122 gabbi-run http://127.0.0.1:8000 < a.yaml

result

`{
  "params": {
    "content": {
      "type": "json",
      "requestInfo": {
        "seqId": "",
        "operator": "csig",
        "requestId": "",
        "caller": "vs-api",
        "requestModule": "csig",
        "requestKey": "mx_ksdionuj3djfk"
      },
      "requestItem": {
        "method": "instance_live_migrate",
        "data": {
          "detail": [
            {
              "schedulerDebug": "on",
              "migrateBandwidth": 20,
              "uuid": 100,
              "compression": {
                "compression": "mt"
              },
              "pauseTime": 122,
              "migrateTimeout": 3600,
              "hostIpList": [
                101
              ]
            }
          ]
        }
      },
      "version": "1.0"
    }
  }
}`

Issue #282, opened 2020-08-19
Verbose misses response body


avkonst:
I have got the following test spec:
```
tests:
-   name: auth
    verbose: all
    url: /api/sessions
    method: POST
    data: "asdsad"
    status: 200
```

which has got data not a proper json on purpose. The response results in 400 instead of 200 code. Verbose is set to all, but it still does not print the response body, although detects non empty content:

```
... #### auth ####
> POST http://localhost:7000/api/sessions
> user-agent: gabbi/1.40.0 (Python urllib3)

< 400 Bad Request
< content-length: 48
< date: Wed, 19 Aug 2020 06:11:24 GMT

✗ gabbi-runner.input_auth

FAIL: gabbi-runner.input_auth
        Traceback (most recent call last):
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 94, in wrapper
            func(self)
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 143, in test_request
            self._run_test()
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 550, in _run_test
            self._assert_response()
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 188, in _assert_response
            self._test_status(self.test_data['status'], self.response['status'])
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 591, in _test_status
            self.assert_in_or_print_output(observed_status, statii)
          File "/usr/lib/python3/dist-packages/gabbi/case.py", line 654, in assert_in_or_print_output
            self.assertIn(expected, iterable)
          File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 417, in assertIn
            self.assertThat(haystack, Contains(needle), message)
          File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 498, in assertThat
            raise mismatch_error
        testtools.matchers._impl.MismatchError: '400' not in ['200']
----------------------------------------------------------------------
```

The expected behavior:
* response body is printed to the stdout alongside the response headers.

cdent:
You're using a version of gabbi that is about 2.5 years old. If you're able to upgrade and try the same test that would be helpful in trying to trace where the problem is. I'm pretty sure the version won't make a difference, but I'd like to remove that variable.

[...]

PR #285, opened 2020-08-20
Display response body when no content-type

cdent:
When the response has no content-type and verbose is `true`
gabbi was not displaying the response body. This is because
when no content-type is known, it was assuming a binary
content-type. Now it assumes text/plain.

This comes with the risk that in some rare cases verbosity
might spew data all over the screen if there is a very big
or very binary response when no content-type is set. With
luck this will encourage servers to send a content-type.

An external test, in the form of test-verbosity.sh is added.
This was one of several ways of verifying the output.

Addresses #282

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -17,6 +17,7 @@ import os
 import sys
 
 import certifi
+import six
 import urllib3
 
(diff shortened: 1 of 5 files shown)


cdent:
This might be too simplistic, or it might be fine. Can't decide.

PR #288, opened 2020-12-16
Be more reslient with pytest fixture start

cdent:
In some cases no tests will be sent to the start method, so instead
of a positional arg, use a keyword and set it to an empty list if
not present.


cdent:
I expect this to fail in some minor ways in travis and intend to only fix those small bits.

A more robust test modernization is required.

PR #291, opened 2020-12-16
Replace Travis CI with GitHub Actions

FND:
> Travis is pretty much dead

see https://github.com/cdent/gabbi/actions

to do:

* [x] fix tests (not sure why pep8 fails 🤷 )
* [ ] ensure that matrix configuration works as expected
* [x] remove this branch from configuration (l. 6)
* [x] update README [badge](https://stackoverflow.com/a/61442205)
* [ ] IRC notifications (see [marketplace](https://github.com/marketplace?type=actions&query=irc) 🤷 )

see [Workflow syntax for GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions) for details


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,48 @@
+name: gabbi tests
```
FND, line 1: not sure whether that's a sensible name - if only because it shows up in URLs:
```
.. image:: https://github.com/cdent/gabbi/workflows/gabbi%20tests/badge.svg
    :target: https://github.com/cdent/gabbi/actions
```

cdent: Is `tests` (by itself) too subtle?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,48 @@
...
+on:
+    push:
+        branches:
+        - master
+        - excision
```
cdent, line 6: does this becomes solely `on: push` since presumably tests should run basically always?

Or is there some kind of more restricted behavior like "only on pull requests" or something that is more the norm?

cdent (commented):
I'll merge a separate patch with pep8 fixes.

.github/workflows/tests.yaml
FND: no, I prefer that too, now that I understand the context of where it shows up (e.g. the badge); will do

.github/workflows/tests.yaml
FND: aye, will do

PR #298, opened 2021-08-13
WIP: Fix JSON content-type matching

### LINKED ISSUE #296: Error when response content-type is application/jsonlines
If the response is `application/jsonlines` the JSON content handler kicks in and attempts to decode the response from json, which fails as it treats the whole doc as one single json object, not lines of json.

This is because the accept method for the handler is checking for a content type that starts with 'application/json'

So the simple fix is straightforward, but it might make better sense to create a jsonlines response handler and see if we can make a way to jsonpath on those lines.

FND:
> this avoids false positives, notably `application/jsonlines`

fixes #296 - maybe? 

I'm not sure this is the right approach either way, plus if we go the RegEx route, we should probably include `+json$` too?

### DIFF
--- a/gabbi/handlers/jsonhandler.py
+++ b/gabbi/handlers/jsonhandler.py
@@ -13,11 +13,15 @@
 """JSON-related content handling."""
 
 import json
+import re
 
 import six
 
 from gabbi.handlers import base
 from gabbi import json_parser
+
+
+CONTENT_TYPE = re.compile('^application/json[^a-zA-Z0-9]*$')
 
 
 class JSONHandler(base.ContentHandler):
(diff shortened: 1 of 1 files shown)


gabbi/handlers/jsonhandler.py
```diff
@@ -13,13 +13,17 @@
...
 import json
+import re
 
 import six
 
 from gabbi.handlers import base
 from gabbi import json_parser
 
 
+CONTENT_TYPE = re.compile('^application/json[^a-zA-Z0-9]*$')
```
cdent, line 24: I guess your reasoning here is "this matches `json` and anything like `json+*`, `json-*`, etc"?

The presupposes that anything `jsonfoo` is not actually json, or is a bad name, which is probably a pretty good assumption. I couldn't find any registered mimetypes that would be an issue.

FND: Yes, the assumption was that "json" followed by alphanumeric characters ain't JSON-as-currently-defined.

No, I didn't think about this in such detail that I explicitly considered `json-[+-]` and such; I just wanted to denylist the above. Seems sensible now though?

cdent:
superseded by #302

thanks for making the ball roll

PR #302, opened 2021-09-02
Constrain how the JSONHandler accepts content-types

### LINKED ISSUE #296: Error when response content-type is application/jsonlines

cdent:
Some formats that look like they might be JSON,
based on the media-type in the content-type header
are not actually valid JSON. This presents problems
with tests trying to decode, but being unable to
do so, causing unexpected failures.

These changes update the accepts method in the
JSONHandler to be more specific and adds tests
to cover it.

As part of this some test which showed that we
accept "application/json-home" as valid JSON have
been removed. That media type is not considered
legitimate. See: https://github.com/mnot/I-D/issues/172

Note that this change is an improvement, but it
is not 100% reliable as there is no good way to
be so because the world is complex and ambiguous,
which is part of why #301 was done: to deal with
gaps.

Thanks to @FND for ideas from #298.

Fixes #296

### DIFF
--- a/gabbi/handlers/jsonhandler.py
+++ b/gabbi/handlers/jsonhandler.py
@@ -38,9 +38,14 @@ class JSONHandler(base.ContentHandler):
 
     @staticmethod
     def accepts(content_type):
-        content_type = content_type.split(';', 1)[0].strip()
+        content_type = content_type.lower()
+        parameters = ''
+        if ';' in content_type:
+            content_type, parameters = content_type.split(';', 1)
+        content_type = content_type.strip()
         return (content_type.endswith('+json') or
-                content_type.startswith('application/json'))
+                content_type == 'application/json'
+                and 'stream' not in parameters)
 
     @classmethod
     def replacer(cls, response_data, match):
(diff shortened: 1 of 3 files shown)


gabbi/handlers/jsonhandler.py
```diff
@@ -38,9 +38,14 @@ class JSONHandler(base.ContentHandler):
...
-        content_type = content_type.split(';', 1)[0].strip()
+        content_type = content_type.lower()
+        parameters = ''
+        if ';' in content_type:
+            content_type, parameters = content_type.split(';', 1)
+        content_type = content_type.strip()
         return (content_type.endswith('+json') or
-                content_type.startswith('application/json'))
+                content_type == 'application/json'
+                and 'stream' not in parameters)
```
cdent, line 48: This should probably be `stream=` not `stream`.

Issue #284, opened 2020-08-20
Consider to provide docker image for gabbi-run


avkonst:
Great tool. I keep using it more and more.
Currently I build my own docker image with gabbi-run.
It would be great if there was an official gabbi-run image provided.

cdent:
Thanks, this is a good idea. And thanks for providing the working sample.

What's probably the best thing to do here is integrate this with the CI so that a new image is built each time a release is tagged. I'll need to do some investigation to make that happen.

cdent:
I've started experimenting with this and I have it basically working, but the interaction is a bit weird.

If you provide the yaml on stdin it can work fine:

```
$ docker  run -i gabbi:latest < /tmp/tests.yaml 
... ✓ gabbi-runner.input_get

----------------------------------------------------------------------
Ran 1 test in 0.085s

OK
```

But if you are used to providing filenames on the arg list it can be a bit more complicated. Do you always use STDIN?

avkonst:
You need to mount the directory with test files. (Also suggest to add `--rm` flag to let docker cleanup the containers when finished).

```
docker  run -it --rm -v /tmp:/proj gabbi:latest -- /proj/tests.yaml 
```

If your test files are in the current directory, it can be mounted to a working directory in the container:

```
docker  run -it --rm -v ./:/wdir -w /wdir gabbi:latest -- tests.yaml 
```

cdent:
That's kind of what I assumed but it feels...clumsy.

I guess if it was scripted it would be no big deal. I'll keep playing with it. Should have something within a few days.

avkonst:
Be aware that deployment with docker is not for the purpose to replace native installation on a development host. It is for the purpose to automate CI tests and simplify trials of a tool without asking people to install python, etc... We use gitlab pipelines. Each job in the pipeline needs a container to run. If we use gabbi for a test job, it has to be an image with gabbi inside of it.

cdent:
Just so you know, I haven't forgotten about this.

PR #304, opened 2021-11-26
Attempt to build and automate container creation

cdent:
This is based on the Dockerfile provided at #284 with some
tweaks.

Makefile is adjusted to allow easy experimentation, but the
primary work here uses github actions.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,43 @@
+name: docker
+
+
+on:
+  push:
+    tags:
+      - '[2-9].[0-9]+.[0-9]+'
+
+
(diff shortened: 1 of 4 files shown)


cdent:
This still needs some documentation, but is otherwise working: https://hub.docker.com/repository/docker/cdent/gabbi

@avkonst if you're still tracking this, if you could a) confirm it works as you expect, b) help me come up with some additions to the docs, that would be great. Sorry it has taken so long. Real life and all that...

avkonst:
hi, no worries. Unfortunately I do not have a project at hands to verify it. Maybe I will have in the future.

PR #306, opened 2022-01-12
Make Gabbi SNI aware. (fixes #305)

### LINKED ISSUE #305: Gabbi isn't SNI-aware
Gabbi isn't compatible with HTTPS servers that require SNI.

scottwallacesh:


cdent:
Thanks for providing this, great addition. I'm surprised this hasn't come up before, but I guess it is because most of the time gabbi is being used to test things where TLS is terminated prior to the app being tested.

Normally I'd want a test of some kind to go along with this, but since there aren't already other test of httpclient, I'll deal with that myself, later.

Look for a new release probably tomorrow. 

Out of curiosity: Are you using gabbi yourself? For anything interesting? It's always useful to know more about what people are doing with it.

scottwallacesh:
> Thanks for providing this, great addition. I'm surprised this hasn't come up before, but I guess it is because most of the time gabbi is being used to test things where TLS is terminated prior to the app being tested.
> 
> Normally I'd want a test of some kind to go along with this, but since there aren't already other test of httpclient, I'll deal with that myself, later.

This is basically a hack to get it working in the way I needed it to.  I thought I'd submit a PR hoping you'd see what I was doing and then do it better.  If you think I've done an okay job as it is, then brilliant!

> Look for a new release probably tomorrow.

Awesome.  No rush, as I say... it was just to get something working for my use case.
 
> Out of curiosity: Are you using gabbi yourself? For anything interesting? It's always useful to know more about what people are doing with it.

The reason I needed SNI working was because we're using Gabbi to test various changes to Akamai's Edge (behaviors, etc.).  Gabbi fits our needs perfectly and we're hoping for some wider adoption once we have a slick testsuite.

cdent:
Whoops, that might have been premature. I had assumed the test actions had run automatically and passed, but looks like they didn't run for some reason. There are some issues, but I'll fix them.

cdent:
It's a bug in wsgi_intercept, which I'll fix

cdent:
2.5.0 is on pypi, and requires 1.9.3 of wsgi-intercept, which is on pypi too

cdent:
So it turns out this change identified a problem in how we are using urllib3.PoolManager, that is probably a bug in urllib3

But it means that 2.5.0 is breaking some CI situations out there in the real world, so I've yanked it on pypi to make it clear that it has a problem. 2.5.1 will come out when the urllib3 issue is resolved. See: #310

PR #308, opened 2022-01-18
Bump urllib3

### LINKED ISSUE #307: urllib>=1.24 is required to use server_hostname
https://github.com/cdent/gabbi/pull/306 introduced usage of the server_hostname argument but this is available only in urllib>=1.24 [1].

[1] https://github.com/urllib3/urllib3/commit/3f2267a2cbe0619ee42249aa13a98a53edbcf07c

We should bump the version in requirements.txt to make sure that the argument is available. Otherwise test run fails with the following error.

```
TypeError: __init__() got an unexpected keyword argument 'server_hostname'
```

kajinamit:
The change 1c43e1c85e2082ca5cfcbc924e6e35e3648329b1 introduced usage
of server_hostname but this is available in urllib3>=1.24.
This change updates the version requirement accordingly.

Fixes #307

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ pbr
 pytest
 six
 PyYAML
-urllib3>=1.11.0
+urllib3>=1.24.0
 certifi
 jsonpath-rw-ext>=1.0.0
 wsgi-intercept>=1.9.3


cdent:
This change is incorporated in #310 which will finally be released today. Thanks very much for your help in finding and fixing these problems.

Issue #311, opened 2022-03-21
Unable to make a relative Content Handler import from the command-line


scottwallacesh:
On the command line, importing a custom Content Handler using a relative path requires manipulation of the `PYTHONPATH` environment variable to add `.` to the list of paths.

Should Gabbi allow relative imports to work out-of-the-box?

e.g. 
```
gabbi-run -r foo.bar:ExampleContentHandler < example.yaml
```
... fails with, `ModuleNotFoundError: No module named 'foo'`.

Updating `PYTHONPATH`...
```
PYTHONPATH=${PYTHONPATH}:. gabbi-run -r foo.bar:ExampleContentHandler < example.yaml
```
... works.

cdent:
/cc @FND 

I had thought we already had a work around for this. Will try to refresh my memory.

scottwallacesh:
We could "detect" a relative import if it was declared as, `.foo.bar:...` and only update the PyPath for the handler initialisation.  Something like... 
```python
def initialize_handlers(response_handlers):
    # Save PyPath
    pypath = sys.path

    custom_response_handlers = []
    handler_objects = []
    for import_path in response_handlers or []:
        # Check if a relative import was provided
        if import_path.startswith('.') and '.' not in sys.path:
            # Allow relative imports
            sys.path.append('.')
        for handler in load_response_handlers(import_path):
            custom_response_handlers.append(handler)
    for handler in handlers.RESPONSE_HANDLERS + custom_response_handlers:
        handler_objects.append(handler())

    # Restore PyPath
    sys.path = pypath
    return handler_objects
```

cdent:
> We could "detect" a relative import if it was declared as, `.foo.bar:...`

That could work but I worry it might be a bit too magical. I suppose it does look enough like a relative import to imply as much.

Any corner cases we're not thinking of?

It's clear enough to document: "If you want to load a module from the local dir prepend a dot..."

scottwallacesh:
> Any corner cases we're not thinking of?

Not that I can think of right now.

scottwallacesh:
Either implementation would work, IMHO. As you say, a `-l` argument would be less "magical".  We're already having to write `-r ...` anyway and it would be simple enough to make it, `-lr ...`.

scottwallacesh:
Oh dear.  Many failed tests.  I'll have to understand how they're run and work out what's going on there.

scottwallacesh:
[...]

PR #312, opened 2022-03-21
Implement a workaround for relative Content Handlers

scottwallacesh:
Fixes, #311

### DIFF
--- a/gabbi/runner.py
+++ b/gabbi/runner.py
@@ -79,7 +79,8 @@ def run():
     host, port, prefix, force_ssl = utils.host_info_from_target(
         args.target, args.prefix)
 
-    handler_objects = initialize_handlers(args.response_handlers)
+    handler_objects = initialize_handlers(
+        args.response_handlers, args.local_handlers)
 
     quiet = args.quiet
     verbosity = args.verbosity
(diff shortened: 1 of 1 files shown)


gabbi/runner.py
```diff
@@ -243,6 +253,13 @@ def _make_argparser():
         help='Custom response handler. Should be an import path of the '
              'form package.module or package.module:class.'
     )
+    parser.add_argument(
+        '-l', '--local-handlers',
+        default=False,
+        dest='local_handlers',
+        action='store_true',
+        help='Response handlers are relative to the current directory.'
```
cdent, line 261: are -> may be

As the change you've done is making it so the can be loaded locally, but they might also be loaded from somewhere else if they are not found on .

scottwallacesh: Updated.

scottwallacesh:
Did you need me to squash the commits so they're clean?

scottwallacesh:
Squashed!

cdent:
I was just getting to merge this and thought of a corner case:

We need to prepend . not append. If we are testing a new version of a handler, and it is local to `.` but there is an already installed version in site-lib, appending . won't help.

Is that right?

Also, if you can think of a reasonable additional test, it could go in gabbi/tests/test_runner.py

scottwallacesh:
Good thought.  Yes, we now prepend rather than append to the path.

PR #319, opened 2022-09-22
Add timeout with a default of 30s.  Fixes #318.

### LINKED ISSUE #318: Tests hang if the server doesn't respond
I cannot discern an appreciable timeout value for requests where the server doesn't respond.  Tests seem to hang.

scottwallacesh:


scottwallacesh:
I am seeing the `tox` failures locally too -- even for the commit before mine, e08fba8.

cdent:
Yeah, this seems like a good addition. I'm surprised it hasn't come up before...

We may wish to consider adding some kind of flag to gabbi-run that would allow setting the value outside of tests, but this is good enough for now.

The tox failures are confusing. If I go back in gabbis history it doesn't get any better so it looks like it is either a urllib3 or python itself change that's messing things up somewhere. I had thought it might be a wsgi-intercept problem, but it appears to be fine. I'll dig into it, but it's going to take some time because I'm on a new computer where I haven't got all the usual versions of python yet, so I'll need to improvise.

I've created #320 to track that separately.

I'll come back to this once that's fixed.

cdent:
If you merge from or rebase to main, this should be good to go and I'll make a new release including it.

cdent:
Thanks very much, 2.8.0 is now out.

PR #322, opened 2023-06-23
Pin urllib3 < 2.0.0

cdent:
As otherwise it breaks lots of things.

Also update to deal with new versions of tox and setuptools
being more strict about modern behaviors.


cdent:
will tidy up commit message via a squash, turned out there was lots of tidying up needed

PR #324, opened 2023-09-23
remove six dependency

a-detiste:
this is a lefotver of Python2 compatibility


cdent:
Thanks for this, will give it a proper look some time a bit later in the week.

Out of curiosity what prompted you to provide the pull request?

a-detiste:
Hi,

I'm reviewing random packages that still depends on six (/ needs a hand).

https://wiki.debian.org/Python3-six-removal

I provided once a -1400/+200 PR to a project,
I learn new things & have some fun.

gabbi/case.py
```diff
@@ -20,16 +20,15 @@
 from collections import OrderedDict
 import copy
 import functools
+import http.cookies
 import os
 import re
 import sys
 import time
 import unittest
+import urllib. parse as urlparse
```
FND, line 29: minor typo?

```suggestion
import urllib.parse as urlparse
```

gabbi/case.py
```diff
@@ -20,16 +20,15 @@
 from collections import OrderedDict
 import copy
 import functools
+import http.cookies
```
FND, line 23: perhaps a little simpler:

```suggestion
from http import cookies
```

(would require corresponding adjustments further down, of course)

gabbi/tests/test_fixtures.py
```diff
@@ -13,10 +13,9 @@
 """Use mocks to confirm that fixtures operate as context managers.
 """
 
+import mock
```
FND, line 16: test failures suggest using this instead:

```suggestion
from unittest import mock
```

FND (commented):
That's a commendable effort!

This all seems sensible upon cursory inspection - though I'm insufficiently contextualized to be a reliable reviewer. I've only noticed a few minor details, plus tests balk at a few stylistic issues (PEP 8).

gabbi/case.py
a-detiste: Of course.

I guess the 3 changes are 2.95% OK so let's do it.

I m typing this on the phone so cannot fix it myself.

gabbi/fixture.py
```diff
@@ -86,7 +84,7 @@ def nest(fixtures):
             contexts.append(enter_func())
             exits.append(exit_func)
         yield contexts
-    except Exception:
+    except Exception as e:
```
cdent, line 87: This is going to fail the pep8 tests, see https://github.com/cdent/gabbi/actions/runs/6331200515/job/17327201194?pr=324

gabbi/tests/__init__.py
```diff
@@ -10,7 +10,3 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-
-import six
-
-six.add_move(six.MovedModule('mock', 'mock', 'unittest.mock'))
```
cdent, line 16: This change causes another pep8 failure. Easiest fix is to remove all the content in the file, leaving it empty.

cdent (changes requested):
Thanks again for doing this, and following up with some fixes. There are a few issues remaining to get the pep8 test to pass.

Locally you can run `tox -epep8` to check for the issues instead of waiting for CI (which clearly I need to adjust permissions on somehow).

If you haven't got time to make the changes let me know and I can do them.

a-detiste:
I find the way the import are sorted completely unnatural.

But now... we have an actual problem to fix(ture) !

gabbi/fixture.py
a-detiste: this was there to made to fixture simpler (simply "raise e"), then got lost in stream of thoughts.

gabbi/tests/__init__.py
a-detiste: done

cdent:
> But now... we have an actual problem to fix(ture) !

Looks like the fix for that is this:

```diff
diff --git a/gabbi/fixture.py b/gabbi/fixture.py
index 4538dee..03b8e10 100644
--- a/gabbi/fixture.py
+++ b/gabbi/fixture.py
@@ -95,4 +95,4 @@ def nest(fixtures):
             except Exception:
                 exc = sys.exc_info()
         if exc != (None, None, None):
-            raise exc[0]
+            raise exc[0](exc[1]).with_traceback(exc[2])
diff --git a/gabbi/tests/test_suite.py b/gabbi/tests/test_suite.py
index d550a4a..e31b0c5 100644
--- a/gabbi/tests/test_suite.py
+++ b/gabbi/tests/test_suite.py
@@ -19,6 +19,7 @@ import unittest
 from gabbi import fixture
 from gabbi import suitemaker

+FIXTURE_METHOD = 'start_fixture'
 VALUE_ERROR = 'value error sentinel'


@@ -54,3 +55,4 @@ class SuiteTest(unittest.TestCase):

         self.assertIn('foo_alpha', str(errored_test))
         self.assertIn(VALUE_ERROR, trace)
+        self.assertIn(FIXTURE_METHOD, trace)
```

If you're happy with that, please apply that and then I'll merge this and make a new release and we can call this a job very well done.

(Note I haven't checked pep8 on these changes yet...)

(Again, if you don't have the time, let me know and I can do it).

gabbi/fixture.py
```diff
@@ -97,4 +95,4 @@ def nest(fixtures):
             except Exception:
                 exc = sys.exc_info()
         if exc != (None, None, None):
-            six.reraise(exc[0], exc[1], exc[2])
+            raise exc[1].with_traceback(exc[2])
```
cdent, line 98: Ah, cool, this is a bit simpler than my suggestion.

cdent (changes requested):
Nice, your solution is a bit simpler than mine, but if you could also add the suggested test change that I made in my proposed diff that would be great. It makes sure that the traceback points to the right location.

gabbi/tests/test_suite.py
```diff
@@ -54,3 +55,4 @@ def test_suite_catches_fixture_fail(self):
 
         self.assertIn('foo_alpha', str(errored_test))
         self.assertIn(VALUE_ERROR, trace)
+        self.assertIn(FIXTURE_METHOD, trace)
```
cdent, line 58: 👍

cdent (approved):

gabbi/tests/test_suite.py
a-detiste: I got the unicode missing rectangle symbol. I had to google it  :-)  ([👍 Pouce Vers Le Haut sur Apple iOS 11.1])

cdent:
Thanks again! 

[2.10.0](https://pypi.org/project/gabbi/) has been released.

PR #325, opened 2024-01-02
Turn on testing for python 3.12

cdent:
This will initially fail until wsgi-intercept is updated for python 3.12 (in progress).


cdent:
see https://github.com/cdent/wsgi-intercept/pull/72 and https://github.com/cdent/wsgi-intercept/pull/73

PR #327, opened 2024-03-02
Add support for regex in response_strings

### LINKED ISSUE #326: Support regex in response_strings
Many of the other test handlers support regex in the value, but `response_strings` does not.

This should probably fixed, but it has the potential to cause errors in existing tests so we may want to consider it as a version 3 to signal the change.

cdent:
Very similar to what is used for headers.

Fixes #326

### DIFF
--- a/gabbi/handlers/core.py
+++ b/gabbi/handlers/core.py
@@ -22,8 +22,20 @@ class StringResponseHandler(base.ResponseHandler):
     test_key_value = []
 
     def action(self, test, expected, value=None):
-        expected = test.replace_template(expected)
-        test.assert_in_or_print_output(expected, test.output)
+        is_regex = (expected.startswith('/') and
+                    expected.endswith('/') and
+                    len(expected) > 1)
+        expected = test.replace_template(expected, escape_regex=is_regex)
+
+        if is_regex:
+            # Trim off /
+            expected = expected[1:-1]
+            test.assertRegex(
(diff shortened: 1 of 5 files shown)


cdent:
Undecided if this is correct/done. Certainly needs docs before being called done, but I though it would be useful to have a look at the initial thing.

gabbi/handlers/core.py
```diff
@@ -22,8 +22,20 @@ class StringResponseHandler(base.ResponseHandler):
     test_key_value = []
 
     def action(self, test, expected, value=None):
-        expected = test.replace_template(expected)
-        test.assert_in_or_print_output(expected, test.output)
+        is_regex = (expected.startswith('/') and
+                    expected.endswith('/') and
+                    len(expected) > 1)
```
FND, line 27: I believe this is now the third occurrence of that particular detection logic, so how about extracting it into a function?

```python
def is_regex(str):
    return re.match("^/.+/$", str) is not None
```

It might additionally do or prepare the `assertRegex` dance beneath.

FND (commented):
lgtm

gabbi/handlers/core.py
cdent: Yeah, good idea, but  I'll stick with starts and ends with.

Which is perhaps ironic, but there you go.

I'll avoid preparing the dance to keep concerns separated.

PR #328, opened 2024-05-30
Uncap urllib3

kajinamit:
The issue with urllib3 >= 2 was already fixed by [1].

[1] https://github.com/urllib3/urllib3/commit/2459900b10f54326745a597cc1cf2bf8a18f0d56

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 pbr
 pytest
 PyYAML
-urllib3>=1.26.9,<2.0.0
+urllib3>=1.26.9
 certifi
 jsonpath-rw-ext>=1.0.0
 wsgi-intercept>=1.13.0


kajinamit:
I might be confused by a different problem we had in the past... Let me check the failing job ...

kajinamit:
Ugh ok so the current problem is different and we have to make wsgi-intercept compatible with urllib3 2 ...

cdent:
Yeah, the situation with wsgi-intercept is big mess, and I'm struggling to figure it out. There is a pull request https://github.com/cdent/wsgi-intercept/pull/76 that fixes it, but it is super fragile and just has the sense of not being the actually correct solution. If you're interested in looking at that and helping me either make it better or decide that it is okay, please feel free. Thanks.

PR #329, opened 2025-01-18
Fix TypeError: exceptions must derive from BaseException

Koalab99:
Hello,

There was an issue with the "raise" line.

I removed it and cleaned a bit, but maybe you want to raise a new error instead of ignoring the WSGIAppError ?

Bye

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -29,6 +29,8 @@ import unittest
 from unittest import result as unitresult
 import urllib.parse as urlparse
 
+import wsgi_intercept
+
 from gabbi import __version__
 from gabbi import exception
 from gabbi.handlers import base
@@ -500,14 +502,19 @@ class HTTPTestCase(unittest.TestCase):
         if 'user-agent' not in (key.lower() for key in headers):
             headers['user-agent'] = "gabbi/%s (Python urllib3)" % __version__
 
-        response, content = self.http.request(
-            url,
-            method=method,
-            headers=headers,
-            body=body,
-            redirect=redirect,
-            timeout=timeout,
-        )
+        try:
+            response, content = self.http.request(
+                url,
+                method=method,
+                headers=headers,
+                body=body,
+                redirect=redirect,
+                timeout=timeout,
+            )
+        except wsgi_intercept.WSGIAppError as exc:
+            # Extract and re-raise the wrapped exception.
+            raise exc.exception_type(exc.exception_value).with_traceback(
+                   exc.traceback) from None
 
         # Set headers and location attributes for follow on requests
         self.response = response
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -29,8 +29,6 @@ import unittest
 from unittest import result as unitresult
 import urllib.parse as urlparse
 
-import wsgi_intercept
-
 from gabbi import __version__
 from gabbi import exception
 from gabbi.handlers import base
@@ -502,19 +500,14 @@ class HTTPTestCase(unittest.TestCase):
         if 'user-agent' not in (key.lower() for key in headers):
             headers['user-agent'] = "gabbi/%s (Python urllib3)" % __version__
 
-        try:
-            response, content = self.http.request(
-                url,
-                method=method,
-                headers=headers,
-                body=body,
-                redirect=redirect,
-                timeout=timeout,
-            )
-        except wsgi_intercept.WSGIAppError as exc:
-            # Extract and re-raise the wrapped exception.
-            raise (exc.exception_type, exc.exception_value,
-                   exc.traceback)
+        response, content = self.http.request(
+            url,
+            method=method,
+            headers=headers,
+            body=body,
+            redirect=redirect,
+            timeout=timeout,
+        )
 
         # Set headers and location attributes for follow on requests
         self.response = response


gabbi/case.py
```diff
@@ -502,19 +500,14 @@ def _run_request(
...
-        try:
-            response, content = self.http.request(
-                url,
-                method=method,
-                headers=headers,
-                body=body,
-                redirect=redirect,
-                timeout=timeout,
-            )
-        except wsgi_intercept.WSGIAppError as exc:
```
cdent, line 514: Won't removing this break what's described in the docstring above:

```
The call to make the request will catch a WSGIAppError from
wsgi_intercept so that the real traceback from a catastrophic
error in the intercepted app can be examined.
```

See 5b2ac224bdac7a341e461ccaf7c48fb6fa36cfed for when it was added. It's entirely possible that was not right, but it was certainly important at the time.

Koalab99:
It was probably important to someone, somewhere, at some point.

But if we trace back this piece of code, a first commit included six to reraise the same error and second commit removed six and added a python error ( da5fa0941aa370cd885052eabb6585d7f3524111 )

I am not a python expert, and it seems very strange to catch an error to reraise it right after without modification. I didn't find the benefits for doing it, maybe you can help me understand how it is important ?

And yes, I should have removed the comment above too. But still, I'm wondering if this is necessary to reraise the same error or not.

cdent:
> I am not a python expert, and it seems very strange to catch an error to reraise it right after without modification. I didn't find the benefits for doing it, maybe you can help me understand how it is important ?

Looking back through the wsgi-intercept code it appears that the code is _not_ rerasiing without modification. WSGIAppError wraps another exception. What the re-raise is doing is raising the wrapped exception, because (as the comment says) we don't want two layers of exceptions when debugging an application we are developing with TDD and testing via gabbi+wsgi-intercept.

What was the original issue you were fixing? That is, under what circumstances are you seeing the TypeError?

Koalab99:
You are right. Thanks for pointing me to the right direction, it was not re-raising the same error. 

To reproduce the problem I had with TypeError:
- In the file: gabbi/tests/gabbits_intercept/self.yaml
- Test named: test exception wrapper
- Comment the "xfail" line temporarily, so tox will show the error

Here is what happened before, when there was an error:
```
Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "./gabbi/case.py", line 506, in _run_request
    response, content = self.http.request(
                        ^^^^^^^^^^^^^^^^^^

      File "./gabbi/httpclient.py", line 40, in request
    response = super(Http, self).request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/request.py", line 81, in request
    return self.request_encode_body(
           ^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/request.py", line 173, in request_encode_body
    return self.urlopen(method, url, **extra_kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/poolmanager.py", line 376, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/connectionpool.py", line 716, in urlopen
    httplib_response = self._make_request(
                       ^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/connectionpool.py", line 468, in _make_request
    six.raise_from(e, None)

      File "<string>", line 3, in raise_from

      File "./.tox/py311/lib/python3.11/site-packages/urllib3/connectionpool.py", line 463, in _make_request
    httplib_response = conn.getresponse()
                       ^^^^^^^^^^^^^^^^^^

      File "/usr/lib/python3.11/http/client.py", line 1370, in getresponse
    response = self.response_class(self.sock, method=self._method)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "/usr/lib/python3.11/http/client.py", line 256, in __init__
    self.fp = sock.makefile("rb")
              ^^^^^^^^^^^^^^^^^^^

      File "./.tox/py311/lib/python3.11/site-packages/wsgi_intercept/__init__.py", line 451, in makefile
    raise WSGIAppError(error, sys.exc_info())

    wsgi_intercept.WSGIAppError: Exception('because you asked me to') at ./gabbi/tests/simple_wsgi.py:58


During handling of the above exception, another exception occurred:


      File "./gabbi/suitemaker.py", line 105, in do_test
    return test_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./gabbi/case.py", line 97, in wrapper
    func(self)

      File "./gabbi/case.py", line 164, in test_request
    self._run_test()

      File "./gabbi/case.py", line 614, in _run_test
    self._run_request(

      File "./gabbi/case.py", line 516, in _run_request
    raise (exc.exception_type, exc.exception_value,

    TypeError: exceptions must derive from BaseException
```

Here is what the new patch does:
```
Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "./gabbi/suitemaker.py", line 105, in do_test
    return test_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./gabbi/case.py", line 97, in wrapper
    func(self)

      File "./gabbi/case.py", line 164, in test_request
    self._run_test()

      File "./gabbi/case.py", line 613, in _run_test
    self._run_request(

      File "./gabbi/case.py", line 516, in _run_request
    raise exc.exception_type(exc.exception_value).with_traceback(exc.traceback) from None

      File "./.tox/py311/lib/python3.11/site-packages/wsgi_intercept/__init__.py", line 449, in makefile
    app_result = self.app(environ, start_response)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "./gabbi/tests/simple_wsgi.py", line 58, in __call__
    raise Exception('because you asked me to')

    Exception: because you asked me to
```
I am not sure it looks perfect, but I think it is more readable and shows the traceback wrapped in WSGIAppError as intended.

cdent:
Ah, cool, thanks for the detailed explanation, this looks good and has the desired behaviour.

cdent:
I'll give it a closer look tomorrow just to be sure but then merge it and do a new release. Thanks again.

cdent (approved):

PR #331, opened 2025-03-11
Add the ability to assert the HTTP version

scottwallacesh:


gabbi/tests/gabbits_intercept/http-version.yaml
```diff
@@ -0,0 +1,7 @@
+# Test, against wsgi-intercept, that the HTTP version can be asserted.
+tests:
+
+- name: http version "header"
+  url: /
+  response_headers:
+    http_version_string: /HTTP\/[0-9]\.[0-9]/
```
cdent, line 7: Can you explain your reasoning behind the choice of the name and the decision to check a full string rather than just the version numbers?

I can see positives and negatives in both ways and hearing your reasoning will probably convince me you're right.

scottwallacesh: Two reasons:
1. When I first looked into this I saw that `urllib3` v2.x provided the value as an integer (`response.version`, as it is in v1.x) and also as a string (`response.version_string`).  I noticed that Gabbi was [already performing some `urllib3` v2.x compatibility tweaks](https://github.com/cdent/gabbi/blob/7377408f2c22ab7620978f231da7a761bc532bdb/gabbi/httpclient.py#L49) and this is an extension of that.
2. Usually the server presents the HTTP version as a string (`HTTP/1.1`, `HTTP/2.0`, etc.) and I would have thought Gabbi's users would be inclined to check it as such.

I'm not 100% convinced the name is good but I thought I'd use something descriptive enough that it wouldn't be seen in the wild.

scottwallacesh: I just realised I may have confused `httplib2` with `urllib3`; probably because that's currently being used by Gabbi.

cdent: I suspect that the right thing to do with this is figure out what the raw data is underneath urllib3 and whether you can get access to that (as a string) and compare against that rather than constructing something from the representation.

Partly because one of the use cases for even being able to check the version is confirming that the server being tested is presenting reasonable data in the response.

I'm not sure of the details of how that would need to be done. May need to dig around under the surface of things.

scottwallacesh: Ahhh... nuts.  The whole point of this change to make assertions that a server will respond with, say, `HTTP/2`.  As Gabbi uses `urllib3` it will only ever make HTTP v1.x requests as `urllib3` doesn't support the newer HTTP protocols.

scottwallacesh:
Closed as there's no point making assertions on the HTTP protocol version as `urllib3` will only ever make HTTP v1.x requests.

cdent:
Oh yeah. Clearly we all need a nice nap because we should have thought of that sooner...

I've started #332 , please discuss there if you have thoughts.

PR #333, opened 2025-03-14
Replace `urllib3` with `httpx`

scottwallacesh:
As per our discussion (#332), here's my naïve attempt at simply replacing `urllib3` with `httpx`.  I've run it with a number of real-world tests and it seems to work.  However, the internal Gabbi tests are not all passing due to the WSGI-Intercept stuff.

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -11,53 +11,53 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import logging
 import os
 import sys
 
-import certifi
-import urllib3
+import httpx
 
 from gabbi.handlers import jsonhandler
 from gabbi import utils
 
-# Disable SSL warnings otherwise tests which process stderr will get
-# extra information.
-urllib3.disable_warnings()
+logging.getLogger('httpx').setLevel(logging.WARNING)
 
 
-class Http(urllib3.PoolManager):
-    """A subclass of the ``urllib3.PoolManager`` to munge the data.
+class Http:
+    """A class to munge the HTTP response.
 
     This transforms the response to look more like what httplib2
     provided when it was used as the HTTP client.
     """
 
+    def __init__(self, **kwargs):
+        self.extensions = {}
+        if 'server_hostname' in kwargs:
+            self.extensions['sni_hostname'] = kwargs['server_hostname']
+        self.client = httpx.Client(verify=kwargs.get('cert_validate', True))
+
     def request(self, absolute_uri, method, body, headers, redirect, timeout):
-        if redirect:
-            retry = urllib3.util.Retry(raise_on_redirect=False, redirect=5)
-        else:
-            retry = urllib3.util.Retry(total=False, redirect=False)
-        response = super(Http, self).request(
-            method,
-            absolute_uri,
-            body=body,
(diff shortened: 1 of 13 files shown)


cdent:
Nice work. I'll look more closely a bit later but for the most part it seems reasonable. I'm uncertain on how best to untangle the wsgi-intercept aspect of the test failures. That will require a bit of thought.

I notice that your editor, or some other part of your process, has reformatted a lot of code that didn't otherwise need to change. That makes it hard to distinguish your actual changes from "cosmetic" changes.

If it is easy for you to do, could you update the PR so it is just the main changes? At some point we should probably `black` the entire codebase, but for a change like this it would be good to have a commit that is clearly only about the main change. Thanks.

If it's a pain, don't worry, we'll work through it.

scottwallacesh:
Yeah, my IDE has called `black`, etc.

Code updated.

gabbi/httpclient.py
```diff
@@ -11,53 +11,58 @@
...
+            response = self.client.request(
+                method=method,
+                url=absolute_uri,
+                headers=headers,
+                content=body,
+                timeout=timeout,
+                follow_redirects=redirect,
+                extensions=self.extensions,
+            )
+        except httpx.ConnectError as error:
```
cdent, line 50: What's the goal behind re-raising this? What happens if you just let it raise?

scottwallacesh: Ahh, this was a little bit of debugging.  I wanted to include the method and URL in the exception string (as, by default, it just says, `invalid hostname` (or something similar)).  We can remove it, if needs be.

cdent: Augmenting the exception with more data is potentially useful but turning it into a RuntimeError is probably not the best choice, might be better to have a custom exception within gabbi that contains the other exception.

The other option is making sure that wherever the exception is eventually caught, somewhere else in gabbi, has more sufficient messaging to provide the necessary context for which test did the failing.

Generally the pattern you want to follow is to not mess with exception this deep, and let them rise up somewhere  else that catches them and makes them meaningful to the observer.

scottwallacesh: Absolutely.  To be honest, I was trying to take on the replacement of the WSGI-Intercept stuff too and this was added for that.  I've removed it.

cdent:
Looks good. I'll see if I can untangle the intercept stuff in a branch below this one.

Once we get that working we should be good to move this forward.

Thanks _very_ much for pushing this along. It is great that gabbi continues to evolve and improve.

cdent:
I'm very very close. Looking good. More tomorrow.

cdent:
I've now got it down to everything working except for the tests in https://github.com/cdent/gabbi/blob/main/gabbi/runner.py . These used wsgi-intercept in a way that can't be replicated with httpx's transport so I'm switching it to use external processes. This is taking longer than expected because there's some complex stuff going on with stdin/stdout/stderr handling that has buffering issues. It's feeling easier to switch to managing subprocesses that untangling the buffers. Will be a few more days.

cdent:
Down to just two failing tests

cdent:
All passing. Will see what needs to be done to target this branch.

PR #337, opened 2025-04-02
Ensure HTTP request headers are sent as strings.

### LINKED ISSUE #336: After moving to `httpx` an integer value in a HTTP request header returns an error
Example test:
```yaml
tests:
  - name: request with integer header value
    url: /foo
    request_headers:
      integer: 13
```

Output:
```text
ERROR: gabbi-runner.test_integer_req_header_request_with_integer_header_value
        Header value must be str or bytes, not <class 'int'>
```

scottwallacesh:
Fixes #336


gabbi/tests/gabbits_runner/test_integer_req_header.yaml
```diff
@@ -0,0 +1,5 @@
+tests:
```
cdent, line 1: Instead of putting this in gabbit_runner and needing an explicit test for it in test_runner, it should work for it to be in `gabbits_intercept` where it will "just run".

scottwallacesh: Test moved.

gabbi/suitemaker.py
```diff
@@ -265,11 +266,24 @@ def test_suite_from_dict(loader, test_base_name, suite_dict, test_directory,
     return file_suite
 
 
+def dict_vals_to_str(in_dict):
+    """`httpx` requires request headers to be strings"""
+    out_dict = {}
+    for key, val in in_dict.items():
+        if not isinstance(val, Sequence):
```
cdent, line 273: Do we have existing support/tests for request_headers being a sequence?

scottwallacesh: No. But I didn't want to encounter any exceptions. Happy to remove/simplify it.

cdent: It's been a while but I think the expected contract here is:

1. request_headers is a dict of scalar values
2. invalid values will cause an exception at the level of the http client that we want to let raise, causing a test error

If I understand how you reached the point of wanting to make this patch, you've encountered a special case where existing tests that were working with a bare numeral in yaml are now failing because httpx is more strict than urllib3 was.

If that's the case we want to continue to allow invalid stuff like sequences to raise to the client and cause whatever error they used to but make a workaround that is very specific to preserving the previous behavior, so what you probably want is something like:

```python
import numbers
for key, value in in_dict.items():
    if isinstance(value, numbers.Number):
        out_dict[key] = str(val)
    else:
       [the rest of it]
```

Furthermore, you probably don't want to be making the change here in test_update. This function is for making sure that the test has the correct base values that all tests have.

A better place for making sure that request_headers have the right format might be in `gabbi/case.py` in `_run_test`, before `_replace_headers_template` is called.

scottwallacesh: Excellent advice, thank you.

I've updated the code.

PR #338, opened 2025-05-15
Fix metadata table in format.rst

pshchelo:
currently the table describing the format of tests metadata is not rendered - neither by github, nor by main doc site - as the last row has 4 columns instead of 3, effectively hiding this info from people reading the documentation.


cdent:
Thanks very much for catching this.

PR #340, opened 2025-05-16
Handle double response formatters with mixed quotes

cdent:
Prior to these changes a test entry like:

  url: $RESPONSE['$.json.path']$RESPONSE['$.json.otherpath']

would experience a parse error as the extracted jsonpath would include portions of the 2nd $RESPONSE string.

This turns out to be because of a greedy regex.

What's interesting and confusing is that the regex, as written, was supposed to be non-greedy already, but it wasn't working correctly.

Further, trying to fix it while maintaining the use of named groups and backreferences did not work.

The result is that we now do slighly more post-processing on slight more complex regexes which might match on arg1 or arg2 (replacers) and case1 or case2 (history indicators) instead of arg and case.

Even with this, there is now one prior test, in
json-left-side.yaml, that had to be changed in order to work properly with the new regexes.

Unless we can come up with something clever, we may need to accept this test change, which is probably fine because the case was very corner.

### DIFF
--- a/gabbi/case.py
+++ b/gabbi/case.py
@@ -277,7 +277,7 @@ class HTTPTestCase(unittest.TestCase):
         Let KeyError raise if variable not present. Capture
         the 'cast' if any.
         """
-        environ_name = match.group('arg')
+        environ_name = match.group('arg1') or match.group('arg2')
         self.cast = match.group('cast')
         return os.environ[environ_name]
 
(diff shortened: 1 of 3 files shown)


cdent:
@FND and @scottwallacesh if you could have a look at this and let me know 👍 or 👎 that would be marvellous. Thanks.

PR #344, opened 2025-06-28
Lazy generation of client (and transport)

kajinamit:
Commit 4d9d3b47e5dc09b7db7a20c2e4b4cf73993a2332 changed when a WSGI app used for intercept is instantiated, and we are no longer able to set up pre-conditions for application using the fixtures hook.

Delay creation of the test client instance (which also instantiate the given WSGI app) to restore the mechanism.

### DIFF
--- a/gabbi/httpclient.py
+++ b/gabbi/httpclient.py
@@ -34,16 +34,28 @@ class Http:
         self.extensions = {}
         if 'server_hostname' in kwargs:
             self.extensions['sni_hostname'] = kwargs['server_hostname']
-        transport = kwargs.get('intercept')
+        self._transport = kwargs.get('intercept')
+        self._version = int(kwargs.get("version", 1))
+        self._script_name = kwargs.get("prefix", "")
+        self._verify = kwargs.get("cert_validate", True)
+        self._client = None
+
+    @property
+    def client(self):
+        if self._client:
+            return self._client
+        transport = self._transport
         if transport:
             transport = httpx.WSGITransport(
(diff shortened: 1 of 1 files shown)


cdent:
> If there's some way to store the lazily-created client on the Suite, that might achieve what you're after?

Note that it would need the client member of the Http class, not the http class itself, as each test needs to have individual control over the `verbose` aspect.

cdent:
In my PR one thing that will need to be checked for correctness is test suites where requests to "real" hosts are mixed with requests to the intercepted apps.

cdent:
> where requests to "real" hosts are mixed with requests to the intercepted apps

It appears to not be working, so the logic probably needs some adjustment.

cdent:
Thinking about this overnight, the code in this patch maintains the same behaviour as >=4.0: Each test has its own transport, thus each test starts its own WSGI app.

As I state above this is probably undesirable. My patch hangs the client on the suite, but then every request uses the WSGITransport meaning any test that is in an intercepted suite that requests an external URL does not work as expected (there's apparently a shortcoming in the existing tests that needs to be fixed as part of addressing this).

When a test URL has no scheme (http or https) then it will get a `self.host` that is a stringified uuid4. Perhaps we could check for this during the http client request and choose between the intercepted client and a non-intercepted client.

Thoughts?

kajinamit:
Sorry I somehow got this out of my mind. OpenStack projects are now attempting to bump gabbi and we are facing this now... I'll get back to this and will try to give some updates. Sorry for long inactivity from my side.

